Issue #13, opened 2024-02-08
Add caching to the service worker + object version checking to update the cache


giac-mysten:
This should speed up loading the page & lower the load on full nodes

Tasks
- [x] Use the [Service Worker's storage API](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#install_and_activate_populating_your_cache) to cache blobs. 
- [x] Check the object version, and update the cache upon change.

PR #86, opened 2024-06-18
feat(portal): fallback to 404.html blob if available

jeka-kiselyov:
Works great overall. Very smooth.

One improvement I'd like: custom 404 blob for a site, something that would be returned for the site if there's nothing found for the path.

Take a look at the sample site I deployed: https://1e1maorotlxkr4feuf7m9jqliu9j12dlr8lxvt28u23f3lur5s.walrus.site/

Everything just works, out of the box, awesome, I was like going to cancel my Heroku plan.

The only thing is that if you navigate to links from the page with history api:

https://1e1maorotlxkr4feuf7m9jqliu9j12dlr8lxvt28u23f3lur5s.walrus.site/ -> https://1e1maorotlxkr4feuf7m9jqliu9j12dlr8lxvt28u23f3lur5s.walrus.site/capsule/0xfd72e8611ed8ebfa2f93acf9640bd8a5096923880feeeae7fabc96d4f6cffab2

it shows the page, but shows 404 if you do refresh. Which is expected, and we can use hash-navigation as a workaround, but would be 100% super if it would support common pattern for SPA - serve specified .html file for any path. (just custom 404 would be great for some cases too).

Adjusted service worker quickly to check if there's '404.html' blob in the root near 'index.html' and serves it on 'Not found' if it's there.

### DIFF
--- a/portal/src/sw.ts
+++ b/portal/src/sw.ts
@@ -325,7 +325,11 @@ async function resolveAndFetchPage(parsedUrl: Path): Promise<Response> {
 async function fetchPage(client: SuiClient, objectId: string, path: string): Promise<Response> {
     const resource = await fetchResource(client, objectId, path);
     if (resource === null || !resource.blob_id) {
-        return siteNotFound();
+        if (path !== '/404.html') {
+            return fetchPage(client, objectId, '/404.html');
+        } else {
+            return siteNotFound();
+        }
     }
 
     console.log("Fetched Resource: ", resource);


mlegner (approved):
Thank you so much for the very first community contribution to Walrus, @jeka-kiselyov! 🥇 

This looks like a great addition to me, but I'll defer to @giac-mysten who is the master behind Walrus Sites.

giac-mysten (approved):
Hey @jeka-kiselyov , many thanks for the first addition! And cool project.
This works! It's something I should have added from the start :) The only thing that is a bit unsatisfactory is that the resolution for a `404` will take almost double the time, which is a bit annoying. 

However, I think this can be fixed in the future by looking at the set of dynamic fields (and checking if there is `/404.html`) and possibly fetching it in parallel. Caching (#13 ) will also help.

jeka-kiselyov:
@mlegner Thank you!

@giac-mysten  Thanks! Yeah, possible option would be to cache everything just on the firrst call of fetchPage with specific objectId, so you already know which blobs are there - and it would be blazingly fast. But... invalidation, that's a tough part. Maybe there's some place where Sui Object version just get updated and it may be used for the quick reference to check in a single RPC call.

It's like the new universe you are building here. Huge space for experiments and improvements. Great.

Issue #92, opened 2024-06-19
Improve site-builder UX


mlegner:
- [X] improve the site-builder defaults
- [X] improve the site builder output -> #65
- [X] add "dry run" to determine the site creation price
- [X] add confirmation for the transaction

Tzal3x:
Hey @mlegner can you elaborate more on the following?
1.  improve the site-builder defaults
2. add confirmation for the transaction

Is this issue currently developed? If so I can start working on it!

mlegner:
Oh, that's a very good question. I think I just took those from https://github.com/MystenLabs/walrus/issues/360, so @giac-mysten might know a bit more...

Regarding the defaults, one nice improvement could be not to require the user to specify a config file if they just want to use the default settings and/or look for the config file in some default locations. (Although I'm not sure that was the original intent.)

In any case, nobody is working on this at the moment.

giac-mysten:
I think "improving the defaults" was about looking for the config files in default locations... Some of this work is already done, but maybe we could have something similar to the `walrus` binary, where the `site-builder` looks for the binary in `~/.config` (@mlegner did we settle the discussion on where the configs should go?) and then tries `./assets/builder-example.yaml`

> add confirmation for the transaction

This is a follow up to the `--dry-run` work. The idea would be for the client to determine the price of creating/updating the site, displaying it to the user, and then asking "publishing/updating will cost XYZ SUI. do you want to proceed with the transaction?". Such that there are no surprises on that front

Issue #88, opened 2024-06-19
Better handling of empty files (Error: the data is empty)


kkomelin:
When I try to publish my frontend app to Walrus Sites, I receive the following error:

```
2024-06-19T07:21:52.356360Z  INFO site_builder: initializing site builder
2024-06-19T07:21:52.356711Z  INFO site_builder: configuration loaded config=Config { portal: "walrus.site", package: 0x514cf7ce2df33b9e2ca69e75bc9645ef38aca67b6f2852992a34e35e9f907f58, general: GeneralArgs { rpc_url: None, wallet: None, walrus_binary: Some("walrus"), walrus_config: None, gas_budget: Some(500000000) } }
2024-06-19T07:21:52.356759Z  INFO site_builder::util: Using wallet configuration from /home/kos/.sui/sui_config/client.yaml
Error: running the command exited with error: 2024-06-19T07:21:59.801087Z  INFO walrus: running in JSON mode
2024-06-19T07:21:59.801167Z  INFO walrus_service::cli_utils: Using Walrus configuration from /home/kos/.walrus/client_config.yaml
2024-06-19T07:21:59.801233Z  INFO walrus_service::cli_utils: Using wallet configuration from /home/kos/.sui/sui_config/client.yaml
2024-06-19T07:21:59.801637Z  INFO walrus_service::cli_utils: Using RPC URL set in wallet configuration
Error: the data is empty
```

kkomelin:
Reproduction demo is in progress. Will publish here shortly

mlegner:
I suspect you have an empty file somewhere. But I agree, we should improve the handling of that and/or print a more useful error.

kkomelin:
The reproduction:

Clone the reproduction repo:
```
git clone -b walrus-deploy https://github.com/kkomelin/sui-dapp-starter
cd sui-dapp-starter
```
Install the deps and run the deploy:
```
pnpm i
pnpm --filter frontend deploy:walrus
```

kkomelin:
If it matters, I've got `robots.txt` empty in the root.

mlegner:
> If it matters, I've got robots.txt empty in the root. But it's intentional.

That makes total sense. Walrus currently doesn't support empty files, so we either need to add support for that or handle it on the Walrus Sites side.

Until we have that, can you just add a comment line to that file as a workaround?

kkomelin:
Yes, good idea. Will try now

kkomelin:
[...]

PR #93, opened 2024-06-19
feat(site-builder): improve site builder output prints and errors

### LINKED ISSUE #65: Provide more INFO output and better errors in site-builder
Currently, the site builder is very quiet. After telling me which wallet is used, it stops printing output. And, depending on the number and size of files, it remains quiet for a long time.

It would be helpful to provide info output like "storing file X" to show the user what is happening. Or at least provide some other progress output.

Also, the errors could be improved. For example for the case where the `walrus` binary cannot be found.

giac-mysten:
A few UI/UX improvements:

- Progress is marked clearly, especially the uploads to Walrus;
- The source of errors in the execution is better identified, especially in the case of
  - empty files;
  - unsupported file extension;
  - missing/inaccessible walrus binary.

Related to #88 
Closes #65 
Contributes to #92


site-builder/src/display.rs
```diff
@@ -0,0 +1,48 @@
...
+            PrintStyledContent(format!("\n{message}\n").green().bold()),
+        )
+        .unwrap();
+    }
+}
+
+pub fn error<S: Display>(message: S) {
+    if cfg!(not(test)) {
+        crossterm::execute!(
+            stdout(),
```
mlegner, line 25: Suggestion: Use `stderr` here.

site-builder/src/site/content.rs
```diff
@@ -121,89 +121,92 @@ pub enum ContentType {
...
+            "woff" => Ok(ContentType::FontWoff),
+            "woff2" => Ok(ContentType::FontWoff2),
+            "xhtml" => Ok(ContentType::ApplicationXhtmlxml),
+            "xls" => Ok(ContentType::ApplicationVndmsexcel),
+            "xlsx" => Ok(ContentType::ApplicationVndopenxmlformatsofficedocumentspreadsheetmlsheet),
+            "xml" => Ok(ContentType::ApplicationXml),
+            "xul" => Ok(ContentType::ApplicationVndmozillaxulxml),
+            "zip" => Ok(ContentType::ApplicationZip),
+            "7z" => Ok(ContentType::ApplicationX7zcompressed),
+            _ => Err(anyhow!(
```
mlegner, line 206: Hint: If you change this to `return Err(...)`, you can remove the `Ok()` from all the other cases and just wrap the whole thing into an `Ok`.

site-builder/src/site/resource.rs
```diff
@@ -330,13 +333,17 @@ impl ResourceManager {
...
     ) -> Result<Option<Resource>> {
-        let content_type = ContentType::from_extension(
+        let extension = full_path.extension().unwrap_or(
             full_path
-                .extension()
-                .ok_or(anyhow!("No extension found for {:?}", full_path))?
-                .to_str()
-                .ok_or(anyhow!("Invalid extension"))?,
+                .file_name()
+                .expect("the path should no terminate in `..`"),
```
mlegner, line 339: Typo:
```suggestion
                .expect("the path should not terminate in `..`"),
```

site-builder/src/site/resource.rs
mlegner, line 1: Remark: I don't see any changes to empty files. Maybe we can at least print a message that we are ignoring them?

mlegner (approved):
Thanks for these UX improvements, @giac-mysten. Just a few minor comments.

site-builder/src/site/resource.rs
giac-mysten: I had kept the error for empty files (see image above). My thinking was that it's better to error out and let the user consider that there is an empty file that cannot be uploaded, and it's up to them to decide what to do.

I can change this to a warning though.

mlegner: Ah, sorry, got confused about this. This works until we have a better support for empty files. 👍

site-builder/src/display.rs
mlegner: I don't see a change here, did you forget to push something?

giac-mysten: I haven't yet pushed  :)

mlegner: Ah, sorry, I was a bit hasty... 😅

PR #114, opened 2024-07-15
refactor(site-builder): deduplicate args for publish and update

Tzal3x:
## Description

Deduplicate the arguments of Publish and Update
commands. 

There was a duplication of some fields that were
shared among the publish and update.

## Changes

The overlapping options were consolidated by
creating a PublishOptions struct that is then
added to both commands.
This new struct is also used in the publish_site
and update_site functions resolving the
"too many arguments" issue.

## Testing

- Rebuild the `site-builder` executable: 
 ```shell  
cargo build --release 
```

- Publish the example site: 
```shell
./target/release/site-builder --config site-builder/assets/builder-example.yaml publish ./examples/snake
``` 

- Update the site using the object id from the above: 
```shell
./target/release/site-builder --config site-builder/assets/builder-example.yaml update ./examples/snake 0x4f5...
```

### DIFF
--- a/site-builder/src/main.rs
+++ b/site-builder/src/main.rs
@@ -228,12 +228,14 @@ async fn run() -> Result<()> {
                 },
         } => {
             publish_site(
-                directory,
-                content_encoding,
-                site_name,
+                PublishOptions {
+                    directory: directory.to_path_buf(),
+                    content_encoding: *content_encoding,
+                    site_name: site_name.to_string(),
+                    epochs: *epochs,
+                    list_directory: *list_directory,
+                },
                 &config,
-                *epochs,
-                *list_directory,
             )
             .await?
         }
--- a/site-builder/src/main.rs
+++ b/site-builder/src/main.rs
@@ -223,7 +223,7 @@ async fn run() -> Result<()> {
         Commands::Publish {
             publish_options,
             site_name,
-        } => publish_site(publish_options.clone(), site_name.to_string(), &config).await?,
+        } => publish_site(publish_options, site_name, &config).await?,
         Commands::Update {
             publish_options,
             object_id,
--- a/site-builder/src/main.rs
+++ b/site-builder/src/main.rs
@@ -219,69 +219,33 @@ async fn run() -> Result<()> {
     config.merge(&args.general);
     tracing::info!(?config, "configuration loaded");
 
-    match &args.command {
+    match args.command {
         Commands::Publish {
-            publish_options:
-                PublishOptions {
-                    directory,
-                    content_encoding,
-                    epochs,
-                    list_directory,
-                },
+            publish_options,
             site_name,
-        } => {
-            publish_site(
-                PublishOptions {
-                    directory: directory.to_path_buf(),
-                    content_encoding: *content_encoding,
-                    epochs: *epochs,
-                    list_directory: *list_directory,
-                },
-                site_name,
-                &config,
-            )
-            .await?
-        }
+        } => publish_site(publish_options.clone(), site_name.to_string(), &config).await?,
         Commands::Update {
-            publish_options:
-                PublishOptions {
-                    directory,
-                    content_encoding,
-                    epochs,
-                    list_directory,
-                },
+            publish_options,
             object_id,
             watch,
             force,
         } => {
-            update_site(
-                PublishOptions {
-                    directory: directory.to_path_buf(),
-                    content_encoding: *content_encoding,
-                    epochs: *epochs,
-                    list_directory: *list_directory,
-                },
-                object_id,
-                &config,
-                *watch,
-                *force,
-            )
-            .await?;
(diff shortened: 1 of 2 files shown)


site-builder/src/publish.rs
```diff
@@ -26,22 +30,34 @@ use crate::{
...
+#[derive(Parser, Debug)]
+pub struct PublishOptions {
+    /// The directory containing the site sources.
+    pub directory: PathBuf,
+    #[clap(short = 'e', long, value_enum, default_value_t = ContentEncoding::PlainText)]
+    /// The encoding for the contents of the site's resources.
+    pub content_encoding: ContentEncoding,
+    /// The name of the site.
+    #[clap(short, long, default_value = "test site")]
+    pub site_name: String,
```
mlegner, line 42: This is currently not used in the update function. I'd suggest either moving this out from the `PublishOptions` struct and to the options of the `Publish` command, or alternatively support changing the site name in an update.

site-builder/src/publish.rs
```diff
@@ -26,22 +30,34 @@ use crate::{
...
+}
+
+pub async fn publish_site(publish_options: PublishOptions, config: &Config) -> Result<()> {
     edit_site(
-        directory,
-        content_encoding,
-        SiteIdentifier::NewSite(site_name.to_owned()),
+        &publish_options.directory,
+        &publish_options.content_encoding,
+        SiteIdentifier::NewSite(publish_options.site_name.to_owned()),
```
mlegner, line 56: Nit: The `to_owned` isn't necessary here.
```suggestion
        SiteIdentifier::NewSite(publish_options.site_name),
```

site-builder/src/main.rs
```diff
@@ -242,40 +218,52 @@ async fn run() -> Result<()> {
...
+                    site_name: site_name.to_string(),
+                    epochs: *epochs,
+                    list_directory: *list_directory,
+                },
                 &config,
-                *epochs,
-                *list_directory,
             )
             .await?
         }
```
mlegner, line 241: Suggestion: Instead of destructuring and reconstructing the `PublishOptions`, let's do one of the following:
1. Change `match &args.command` to `match args.command` and then use `publish_options` directly:
   ```suggestion
           Commands::Publish { publish_options } => publish_site(publish_options, &config).await?,

   ```
   This requires changing some lines below (removing some `*`, `to_string`, ..., and adding some `&`).
2. Add a `#[derive(Clone)]` to `PublishOptions` (which we should anyway have) and change this to
   ```rust
           Commands::Publish { publish_options } => {
               publish_site(publish_options.clone(), &config).await?
           }
   ```

The same applies to the `Update` command below.

mlegner (changes requested):
Thank you so much for your PR, @Tzal3x!

This looks good to me, just a few minor suggestions.

site-builder/src/main.rs
Tzal3x: Got it! Why do we need to use this though?
> Change match &args.command to match args.command

mlegner: The `publish_site` takes `publish_options` as an owned object; if we do `match &args.command` we only have it as a reference. Of course, an additional option would be to change the argument of `publish_site` to `publish_options: &PublishOptions`.

site-builder/src/main.rs
```diff
@@ -240,59 +219,33 @@ async fn run() -> Result<()> {
...
-                directory,
-                content_encoding,
-                site_name,
-                &config,
-                *epochs,
-                *list_directory,
-            )
-            .await?
-        }
+        } => publish_site(publish_options.clone(), site_name.to_string(), &config).await?,
```
mlegner, line 226: Nit:
```suggestion
        } => publish_site(publish_options, site_name, &config).await?,
```

site-builder/src/main.rs
```diff
@@ -240,59 +219,33 @@ async fn run() -> Result<()> {
...
-                content_encoding,
-                object_id,
-                &config,
-                *watch,
-                *epochs,
-                *force,
-                *list_directory,
-            )
-            .await?;
+            update_site(publish_options.clone(), &object_id, &config, watch, force).await?;
```
mlegner, line 233: Nit:
```suggestion
            update_site(publish_options, &object_id, &config, watch, force).await?;
```

mlegner (approved):
Thanks for the changes, LGTM! 🥇 

Just two nits, then we can get this merged. 🙂

site-builder/src/main.rs
mlegner: Ah, I just realized that my suggestion above was a bit unclear. I meant to do one of the two things, not both (although both together also works). 😇

Tzal3x:
> Thanks for the changes, LGTM! 🥇 
> 
> 
> 
> Just two nits, then we can get this merged. 🙂 

I committed them both directly from gh 👌

PR #117, opened 2024-07-18
feat(site-builder): on publish raise an error when a file is empty

Tzal3x:
So far the `site-builder` was silently ignoring empty files.

This PR adds a check that raises an error when an empty file is included in the directory that is to be published.
Also, the check should be done before trying to create a walrus blob id.   

i.e. 
```shell
❯ ./target/release/site-builder --config site-builder/assets/builder-example.yaml publish examples/snake/

...
Parsing the directory examples/snake/ and locally computing blob IDs ...
Error during execution
Error: error while reading resource `examples/snake/empty.txt`

Caused by:
    0: Walrus sites does not support empty files.
    1: The file examples/snake/empty.txt is empty.
```

### DIFF
--- a/site-builder/src/site/resource.rs
+++ b/site-builder/src/site/resource.rs
@@ -347,10 +347,11 @@ impl ResourceManager {
 
         let plain_content = std::fs::read(full_path)?;
         if plain_content.is_empty() {
-            let error_message = format!("The file {} is empty.", full_path.to_string_lossy());
-            return Err(
-                anyhow!(error_message).context("Walrus does not support empty files.")
+            let error_message = format!(
+                "The file {} is empty. Walrus does not support empty files.",
+                full_path.to_string_lossy()
             );
+            return Err(anyhow!(error_message));
         }
         // TODO(giac): this could be (i) async; (ii) pre configured with the number of shards to
         //     avoid chain interaction (maybe after adding `info` to the JSON commands).
--- a/site-builder/src/site/resource.rs
+++ b/site-builder/src/site/resource.rs
@@ -347,11 +347,10 @@ impl ResourceManager {
 
         let plain_content = std::fs::read(full_path)?;
         if plain_content.is_empty() {
-            let error_message = format!(
-                "The file {} is empty.",
-                full_path.to_string_lossy()
+            let error_message = format!("The file {} is empty.", full_path.to_string_lossy());
+            return Err(
+                anyhow!(error_message).context("Walrus sites does not support empty files.")
             );
-            return Err(anyhow!(error_message).context("Walrus sites does not support empty files."));
         }
         // TODO(giac): this could be (i) async; (ii) pre configured with the number of shards to
         //     avoid chain interaction (maybe after adding `info` to the JSON commands).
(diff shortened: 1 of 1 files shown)


site-builder/src/site/resource.rs
```diff
@@ -344,15 +344,17 @@ impl ResourceManager {
...
-        let output = self.walrus.blob_id(full_path.to_owned(), None)?;
-        // Need to check the size to avoid calling encode on empty files.
+
         let plain_content = std::fs::read(full_path)?;
         if plain_content.is_empty() {
-            // We are ignoring empty files.
-            return Ok(None);
+            let error_message = format!("The file {} is empty.", full_path.to_string_lossy());
+            return Err(
+                anyhow!(error_message).context("Walrus sites does not support empty files.")
```
giac-mysten, line 352: Nit: technically, it is Walrus that does not support empty files.

```suggestion
                anyhow!(error_message).context("Walrus does not support empty files.")
```

giac-mysten, line 352: Suggestion: AFAIU, `context` is added when propagating errors. Since here you are creating the error, you could have the full message saying `The file {} is empty; Walrus does not support..`. Or is there another reason to have the two-level error message?

giac-mysten (changes requested):
Thanks for the change, @Tzal3x ! One general question: According to the [discussion in the issue](https://github.com/MystenLabs/walrus-sites/issues/88#issuecomment-2178056179), empty files should have been removed or "filled" somehow, with a warning message printed for the user. Here we error out instead. Any specific reason for this UI decision? I'm not sure what is best tbh.

site-builder/src/site/resource.rs
Tzal3x: I did this actually because it looked readable to me, but on second thought context is better to be use for propagating errors as you suggest.

site-builder/src/site/resource.rs
```diff
@@ -344,15 +344,18 @@ impl ResourceManager {
...
-        //     avoid chain interaction (maybe after adding `info` to the JSON commands).
-        let output = self.walrus.blob_id(full_path.to_owned(), None)?;
-        // Need to check the size to avoid calling encode on empty files.
+
         let plain_content = std::fs::read(full_path)?;
         if plain_content.is_empty() {
-            // We are ignoring empty files.
-            return Ok(None);
+            let error_message = format!(
+                "The file {} is empty. Walrus does not support empty files.",
```
giac-mysten, line 351: Nit: I just noticed that here and in other parts of walrus sites we use capitalized errors. However, it would be best to follow the stdlib convention (which I believe we use in walrus too) of having error messages as "concise lowercase sentences without trailing punctuation:" (see [docs](https://doc.rust-lang.org/stable/std/error/trait.Error.html))
```suggestion
                "the file {} is empty; Walrus does not support empty files",
```

giac-mysten (approved):
Thanks again @Tzal3x ! sorry for the pedantic nit 😬

Tzal3x:
> Thanks again @Tzal3x ! sorry for the pedantic nit 😬

Any feedback is always valuable 😁

PR #120, opened 2024-07-19
feat(service-worker): track seen resources to avoid infinite redirect loops

Tzal3x:
When recursively redirecting to the object ids of a Display, keep track of the "visited" object ids to avoid 
infinite loops. Source: https://github.com/MystenLabs/walrus-sites/pull/58#discussion_r1635253806

Also refactored the return type of the fetchResource function It would return `null` for cases where the resource would not be able to load. 

This is not ideal, since we need to be more expressive with our responses (we introduce another error of INFINITE_LOOPS 508 and 310 for maximum number of redirects reached). 

Using an http status codes enum enables us to propagate the nature of those errors to the response of `fetchPage`.

### DIFF
--- a/portal/src/http_status_codes.ts
+++ b/portal/src/http_status_codes.ts
@@ -3,16 +3,7 @@
 
 // Purpose: Contains the enumeration of HTTP status codes.
 export enum HttpStatusCodes {
-    OK = 200,
-    CREATED = 201,
-    NO_CONTENT = 204,
-    MOVED_PERMANENTLY = 301,
-    FOUND = 302,
     TOO_MANY_REDIRECTS = 310,
-    BAD_REQUEST = 400,
-    UNAUTHORIZED = 401,
-    FORBIDDEN = 403,
     NOT_FOUND = 404,
-    INTERNAL_SERVER_ERROR = 500,
     LOOP_DETECTED = 508
 }
--- a/portal/src/http_status_codes.ts
+++ b/portal/src/http_status_codes.ts
@@ -1,0 +1,18 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Purpose: Contains the enumeration of HTTP status codes.
+export enum HttpStatusCodes {
+    OK = 200,
+    CREATED = 201,
+    NO_CONTENT = 204,
+    MOVED_PERMANENTLY = 301,
+    FOUND = 302,
+    TOO_MANY_REDIRECTS = 310,
+    BAD_REQUEST = 400,
+    UNAUTHORIZED = 401,
+    FORBIDDEN = 403,
+    NOT_FOUND = 404,
+    INTERNAL_SERVER_ERROR = 500,
+    LOOP_DETECTED = 508
+}
--- a/portal/src/walrus-sites-sw.ts
+++ b/portal/src/walrus-sites-sw.ts
@@ -322,7 +322,7 @@ async function resolveAndFetchPage(parsedUrl: Path): Promise<Response> {
  * Fetches a page.
  */
 async function fetchPage(client: SuiClient, objectId: string, path: string): Promise<Response> {
-    const resource = await fetchResource(client, objectId, path);
+    const resource = await fetchResource(client, objectId, path, new Set<string>);
     if (resource === null || !resource.blob_id) {
         if (path !== '/404.html') {
             return fetchPage(client, objectId, '/404.html');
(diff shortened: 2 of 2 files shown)


portal/src/walrus-sites-sw.ts
```diff
@@ -365,16 +385,25 @@ async function fetchPage(client: SuiClient, objectId: string, path: string): Pro
...
  */
 async function fetchResource(
     client: SuiClient,
     objectId: string,
     path: string,
-    depth: number = 0
-): Promise<Resource | null> {
+    seenResources: Set<string>,
+    depth: number = 0,
+): Promise<Resource | HttpStatusCodes> {
```
giac-mysten, line 398: Praise: I like this way of returning status

portal/src/http_status_codes.ts
```diff
@@ -0,0 +1,18 @@
...
+    MOVED_PERMANENTLY = 301,
+    FOUND = 302,
+    TOO_MANY_REDIRECTS = 310,
+    BAD_REQUEST = 400,
+    UNAUTHORIZED = 401,
+    FORBIDDEN = 403,
+    NOT_FOUND = 404,
+    INTERNAL_SERVER_ERROR = 500,
+    LOOP_DETECTED = 508
+}
```
giac-mysten, line 18: Suggestion: I like this way of presenting status codes. However, we are only using (afaics) `LOOP_DETECTED, TOO_MANY_REDIRECTS, NOT_FOUND`. I would propose to adhere to @mlegner 's YAGNI principle, and remove  the others until we need them.

```suggestion
export enum HttpStatusCodes {
    TOO_MANY_REDIRECTS = 310,
    NOT_FOUND = 404,
    LOOP_DETECTED = 508
}
```

portal/src/walrus-sites-sw.ts
```diff
@@ -322,31 +336,37 @@ async function resolveAndFetchPage(parsedUrl: Path): Promise<Response> {
  * Fetches a page.
  */
 async function fetchPage(client: SuiClient, objectId: string, path: string): Promise<Response> {
-    const resource = await fetchResource(client, objectId, path);
-    if (resource === null || !resource.blob_id) {
+    const result = await fetchResource(client, objectId, path, new Set<string>);
+    const unsuccessfulResourceFetch = !isResource(result)
+    if (unsuccessfulResourceFetch) {
+        const httpStatus = result as number;
+        return new Response("Unable to fetch the site resource.", { status: result });
```
giac-mysten, line 343: Question: How is this rendered? Should this be using `Response404` to show the error nicely?

giac-mysten (changes requested):

giac-mysten:
Thanks for the PR @Tzal3x ! just a few questions and minor fixes from my side

portal/src/walrus-sites-sw.ts
```diff
@@ -38,6 +39,19 @@ type Resource = {
     blob_id: string;
 };
 
+/**
+ * Type guard for the Resource type.
+*/
+function isResource(obj: any): obj is Resource {
```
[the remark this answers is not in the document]
Tzal3x: Thanks! Type guards are a very useful feature of typescript. It's a neat way to check the type of an object.

portal/src/walrus-sites-sw.ts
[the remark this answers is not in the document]
Tzal3x: Ah yes, good catch 👌

portal/src/http_status_codes.ts
Tzal3x: Fair! 🙆‍♂️

portal/src/walrus-sites-sw.ts
Tzal3x: Right now it doesn't get rendered as an html page, so yeah this is not pretty by the user's perspective.

The reason I didn't use Response404 is because the error status returned by `fetchResource` can be different than this (e.g. LOOP_DETECTED, TOO_MANY_REDIRECTS etc). 

I think it would be nice having another template for `300` errors and `500` errors. 
We can make an issue about this and implement it on another PR.

giac-mysten: That makes sense. We could maybe rename `Response404` to `ResponseError`, and have each error code and a message passed to it, s.t. both the error code and the message are printed. Wdyt?

giac-mysten (approved):
Thanks for the changes, good stuff!

portal/src/walrus-sites-sw.ts
Tzal3x: Sounds good to me!

Tzal3x: Created this issue: https://github.com/MystenLabs/walrus-sites/issues/122

Issue #32, opened 2024-06-05
Allow SuiNs subdomains to work with the service worker


giac-mysten:
for example `nft1.myproject.walrus.site`. 
This also allows to programmatically add site names on mint.

Tzal3x:
Currently suins names don't support dots: 
![image](https://github.com/user-attachments/assets/0be35b52-188d-43e5-9bd4-e94141297b9d)
You can test it out here: https://testnet.suins.io/account/my-names#your_name

Issue #26, opened 2024-02-13
Add support for longer "tlds" such as `co.uk`


giac-mysten:
currently the portal enforces that the host is in the form `subdomain.domain.tld`, i.e., `host.split(".").length === 3`. Therefore, it won't work if the portal is registered at `portal.co.uk`

PR #121, opened 2024-07-19
feat(portal): Add support for tlds and suins subnames

Tzal3x:
We are currently supporting only a pattern of domains like this `0x123.walrus.sites`.

This PR makes it possible to:
- Parse more complex domains: `0x123.localhost:8080`, `0x123.walrus.co.uk`
- Support suins subnames: `subsubname.subname.suinsname.walrus.sites`
- Includes some unit tests so that we can have more confidence on the function's results. Gradually we can extend our code coverage. 

Connected issues:
- Closes https://github.com/MystenLabs/walrus-sites/issues/32
- Closes https://github.com/MystenLabs/walrus-sites/issues/26

### DIFF
--- a/.github/workflows/portal-tests.yml
+++ b/.github/workflows/portal-tests.yml
@@ -35,5 +35,4 @@ jobs:
         run: pnpm install
 
       - name: Run tests
-        working-directory: portal/
-        run: echo $(ls) && pnpm run test
+        run: cd portal && pnpm run test
--- a/.github/workflows/portal-tests.yml
+++ b/.github/workflows/portal-tests.yml
@@ -1,0 +1,39 @@
+name: Portal Unit Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Change directory to portal
+        run: cd portal
+
+      - uses: pnpm/action-setup@v4
+        name: Install pnpm
+        with:
+          version: 8
+          run_install: false
+
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install dependencies
+        working-directory: portal
+        run: pnpm install
+
+      - name: Run tests
+        working-directory: portal/unit_tests
+        run: ls ../src/helpers && pnpm run test
--- a/.github/workflows/portal-tests.yml
+++ b/.github/workflows/portal-tests.yml
@@ -16,23 +16,17 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Change directory to portal
-        run: cd portal
-
       - uses: pnpm/action-setup@v4
         name: Install pnpm
         with:
+          working-directory: portal
           version: 8
-          run_install: false
+          run_install: true
 
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
           node-version: 20
-
-      - name: Install dependencies
-        working-directory: portal
-        run: pnpm install
 
       - name: Run tests
         run: cd portal && pnpm run test
--- a/.github/workflows/portal-tests.yml
+++ b/.github/workflows/portal-tests.yml
@@ -35,5 +35,5 @@ jobs:
         run: pnpm install
 
       - name: Run tests
-        working-directory: portal/unit_tests
-        run: ls ../src/helpers && pnpm run test
+        working-directory: portal/
+        run: echo $(ls) && pnpm run test
--- a/portal/package.json
+++ b/portal/package.json
@@ -3,7 +3,7 @@
 		"@mysten/bcs": "^1.0.2",
 		"@mysten/sui": "^1.1.2",
 		"base-x": "^4.0.0",
-		"parse-domain": "^8.0.2",
+		"parse-domain": "8.0.2",
 		"ts-loader": "^9.5.1",
 		"tsc": "^2.0.4",
 		"typescript": "^5.4.5"
--- a/portal/package.json
+++ b/portal/package.json
@@ -3,6 +3,7 @@
 		"@mysten/bcs": "^1.0.2",
 		"@mysten/sui": "^1.1.2",
 		"base-x": "^4.0.0",
+		"parse-domain": "^8.0.2",
 		"ts-loader": "^9.5.1",
 		"tsc": "^2.0.4",
 		"typescript": "^5.4.5"
--- a/portal/package.json
+++ b/portal/package.json
@@ -11,6 +11,7 @@
 		"copy-webpack-plugin": "^12.0.2",
 		"css-minimizer-webpack-plugin": "^6.0.0",
 		"html-minimizer-webpack-plugin": "^5.0.0",
+		"vitest": "^2.0.3",
 		"webpack": "^5.92.0",
 		"webpack-cli": "^5.1.4",
 		"webpack-dev-server": "^4.15.2",
@@ -22,6 +23,7 @@
 		"build": "pnpm build:prod",
 		"dev": "pnpm build:dev",
 		"build:dev": "webpack --config webpack.config.dev.js",
-		"build:prod": "webpack --no-watch --config webpack.config.prod.js"
+		"build:prod": "webpack --no-watch --config webpack.config.prod.js",
+		"test": "vitest"
 	}
 }
--- a/portal/src/helpers/domain_parsing.ts
+++ b/portal/src/helpers/domain_parsing.ts
@@ -9,19 +9,20 @@ import { Path } from "../types/index";
  * @param orig_url The URL to extract the domain from. e.g. "https://example.com"
  * @returns The domain of the URL. e.g. "example.com"
  */
-export function getDomain(orig_url: string): string {
-    const urlStripped = stripProtocolAndPort(orig_url);
-    const parsed = parseDomain(urlStripped);
-    if (parsed.type === ParseResultType.Listed) {
-        const domain = parsed.domain + "." + parsed.topLevelDomains.join(".");
-        return domain;
-    } else if (parsed.type === ParseResultType.Reserved) {
-        return parsed.labels[parsed.labels.length - 1];
-    } else {
-        console.error("Error while parsing domain name:", parsed);
-        throw new Error("Error while parsing domain name");
-    }
-}
+ export function getDomain(orig_url: string): string {
+     // const urlStripped = stripProtocolAndPort(orig_url);
+     const urlStripped = new URL(orig_url).hostname;
+     const parsed = parseDomain(urlStripped);
+     if (parsed.type === ParseResultType.Listed) {
+         const domain = parsed.domain + "." + parsed.topLevelDomains.join(".");
+         return domain;
+     } else if (parsed.type === ParseResultType.Reserved) {
+         return parsed.labels[parsed.labels.length - 1];
+     } else {
+         console.error("Error while parsing domain name:", parsed);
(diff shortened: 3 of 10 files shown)


portal/unit_tests/domain_parsing.test.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, it, expect, test } from 'vitest'
+import { getDomain } from '../src/helpers/domain_parsing.ts'
+
+const testCases: [string, string][] = [
+    ['https://example.com', 'example.com'],
+    ['https://suinsname.localhost:8080', 'localhost'],
+    ['https://suinsname.subname.localhost:8080', 'localhost'],
```
mlegner, line 10: Important: Following standard DNS conventions, this should rather be `subname.suinsname.[...]`, also in the other subname examples.

Tzal3x: You mean it should be `['https://suinsname.subname.localhost:8080', 'suinsname.subname.localhost']`?

Tzal3x:
I still need to check if the suins subnames work as expected. 
Converting this back to draft PR until then.

portal/unit_tests/domain_parsing.test.ts
mlegner: No, I mean that we generally want the `subname` to be before the `suinsname`.

Tzal3x: When I am hitting this url http://flatland.localhost:8080/#0x49de81ca6b68b5956a26ad275e0a0e7fb3e7d0165134bdd713e4719c9be8522b I see on the console the following: `Portal domain and request domain:  localhost localhost`. 
So I presume we should only parse the domain and not the suins + subnames part of the url in this function.

mlegner: > So I presume we should only parse the domain and not the suins + subnames part of the url in this function.

Yes, I agree. But for any follow-up, we need to ensure that we parse the SuiNS name and any subnames correctly as well. And domain names start with the most specific name and then up to more general. So the `subname` needs to come before the `suinsname`. So, for example, `subsubname.subname.suinsname.somedomain.sometld`.

Tzal3x: Ok clear! I will update this, thanks @mlegner!

.github/workflows/portal-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+          version: 8
+          run_install: false
+
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install dependencies
+        working-directory: portal
```
giac-mysten, line 34: Question: why `working-directory: portal` here, while elsewhere there is `cd portal`?

portal/package.json
```diff
@@ -3,6 +3,7 @@
 		"@mysten/bcs": "^1.0.2",
 		"@mysten/sui": "^1.1.2",
 		"base-x": "^4.0.0",
+		"parse-domain": "^8.0.2",
```
giac-mysten, line 6: Additional Remark: [This notice](https://github.com/peerigon/parse-domain?tab=readme-ov-file#updates) on the package seems very important. Do we have an automated way to update this "everytime you start or build your application"?

giac-mysten (changes requested):
Hey @Tzal3x , thanks for the changes! There are some issues with the `parse-domain` library I'd like to discuss more in depth :)

.github/workflows/portal-tests.yml
Tzal3x: Ah that was a remnant while I was testing the difference between the two. 
`working-directory` should be used because it has no side effects on next steps compared to `cd`.

portal/src/helpers/domain_parsing.ts
```diff
@@ -0,0 +1,55 @@
...
+            path: url.pathname == "/" ? "/index.html" : removeLastSlash(url.pathname)
+        } as Path;
+    } else if ( parsed.type === ParseResultType.Reserved) {
+        return {
+            subdomain: parsed.labels.slice(0, parsed.labels.length-1).join('.'),
+            path: url.pathname == "/" ? "/index.html" : removeLastSlash(url.pathname)
+        } as Path;
+    }
+    return null;
+}
```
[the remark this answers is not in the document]
Tzal3x: Yes I considered doing this but I ended up preferring two functions instead of one, because there is better separation of concerns. Even though we repeat a little bit of code, it's a good trade-off for readability imho. 

If you disagree, I can merge them into one, no problem. 👍

portal/src/walrus-sites-sw.ts
```diff
@@ -6,7 +6,9 @@ import * as baseX from "base-x";
 import { fromB64, fromHEX, isValidSuiObjectId, isValidSuiAddress, toHEX } from "@mysten/sui/utils";
 import { AGGREGATOR, SITE_PACKAGE, SITE_NAMES, NETWORK, MAX_REDIRECT_DEPTH } from "./constants";
 import { bcs, BcsType } from "@mysten/bcs";
-import template_404 from "../static/404-page.template.html";
+import template_404 from "@static/404-page.template.html";
```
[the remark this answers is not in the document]
Tzal3x: It's a convention of naming paths in `tsconfig.json`. 
[TS path definitions](https://www.typescriptlang.org/tsconfig/#paths) are a way of mapping relative paths to predetermined ones. i.e. `"@static": "../static"`. 
I added it so that it looks nicer, no other reason.

.github/workflows/portal-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@v4
+
+      - name: Change directory to portal
+        run: cd portal
+
+      - uses: pnpm/action-setup@v4
+        name: Install pnpm
+        with:
+          version: 8
+          run_install: false
```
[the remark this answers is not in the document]
Tzal3x: That is because I wanted to separate those steps of `pnpm setup` and `pnpm install. 
I can change it to `run_install: true`, it won't make a significant difference.

portal/package.json
Tzal3x: That's true, I should include `npx parse-domain-update` to the build and test steps.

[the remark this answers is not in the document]
Tzal3x: > we tend to only import dependencies that (i) we know are stable and well maintained, and (ii) do not offer functionality that we could quickly code and maintain ourselves

I think that it covers our use case without introducing too much complexity, so no redundancy is introduced. 

Apart from the amount of stars, I also looked at the weekly downloads (~150K), which looks good to me. 
Plus it is battle-tested from other users so it would also be likely that we have a bug in our implementation. 
At its' core, it identifies parts of the URL using the https://publicsuffix.org/ and extracts parts of the code. 

To me it looks fine as is, but would you prefer an implementation from scratch?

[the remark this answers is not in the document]
Tzal3x: You are right, that would break, because I was thinking of `walrus.site` as a domain of a portal `walrus` and with top level domain `site`. 🤔

To test that it fails I tested the `getDomain` with input `flatland.site`. The expected value would be `site`, but I received `flatland.site`. The same would be for `walrus.site` if it's listed. The domain will show up as `flatland.walrus.site`.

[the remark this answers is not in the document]
Tzal3x: In this case, would it make sense supporting domains like this though? i.e. `https://flatland.com` flatland is the suinsname, but what would the portal be?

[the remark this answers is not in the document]
Tzal3x: Moved this one to issue https://github.com/MystenLabs/walrus-sites/issues/124 after our call.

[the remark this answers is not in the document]
Tzal3x: Moved this one to issue https://github.com/MystenLabs/walrus-sites/issues/124 after our call.

.github/workflows/portal-tests.yml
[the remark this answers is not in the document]
Tzal3x: Alright I included the `run_install: true` to be more concise.

portal/package.json
Tzal3x: Quoting parse-domain [README](https://github.com/peerigon/parse-domain#readme):

>This package comes with a prebuilt list that has been downloaded at the time of npm publish. 

Therefore, if we use a fixed version of the `parse-domain`  (e.g. `8.0.2`, the current one) the code will still be functional even if we do add the walrus.site to the suffix list.

.github/workflows/portal-tests.yml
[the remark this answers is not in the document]
giac-mysten: Nice, thanks! was just wondering what the difference was, but this seems to work well.

.github/workflows/portal-tests.yml
giac-mysten: Ok, up to you if this needs further investigation, or if the result works.

portal/package.json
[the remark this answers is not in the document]
giac-mysten: Closing after the discussion on this that we had offline. The package looks good, and we are tracking #124

giac-mysten: Perfect!

portal/src/helpers/domain_parsing.ts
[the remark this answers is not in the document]
giac-mysten: My concern is that when we will need to make the change in #124 we will need to remember to touch both of these functions. Therefore, I would be in favor of not duplicating the logic.

If you like the two top level functions, maybe we can keep them and have them call internally another `splitUrl` functions that returns `{subdomain: "subsub.sub", domain: "example.com", path: "/some/path.html"}`? Such that the URL processing logic is as centralized as possible.

portal/src/walrus-sites-sw.ts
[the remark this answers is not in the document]
giac-mysten: thanks for the explanation!

portal/src/helpers/domain_parsing.ts
```diff
@@ -0,0 +1,56 @@
...
+        throw new Error("Error while parsing domain name");
+    }
+}
+
+/**
+* Given a URL, returns the subdomain and path.
+* @param url e.g. "https://subname.name.walrus.site/"
+* @returns Path object e.g. { subdomain: "subname.name", path: "/index.html"}
+*/
+
```
giac-mysten, line 30: Nit
```suggestion
```

giac-mysten (approved):
Thanks @Tzal3x ! just left one comment open about merging the two functions. Please go ahead and merge if you think it is not necessary 👍

portal/src/helpers/domain_parsing.ts
```diff
@@ -28,22 +19,41 @@ export function getDomain(url: URL): string {
...
-        } as Path;
+            domain: null,
+            path: null
+        }
     }
-    return null;
+
+    return {
+        domain,
+        path: {
```
giac-mysten, line 49: Nit: having `path` as a name nested twice is a bit odd... I think I made a mistake initially in calling the class `Path` (it was for something else at the beginning, and the name stuck). Can we have the `UrlExtract` struct be flat, and then have an instance method that returns the `Path` struct? (also maybe rename `Path`).

giac-mysten (approved):
Just a nit; happy for you to merge as is tho!

portal/src/helpers/domain_parsing.ts
Tzal3x: I wouldn't disagree with this, I will rename the `Path` to something else, I had this in mind too.
 
On the other hand, since `UrlExtract` is a type definition and not a struct, it's not possible to have getter methods in it to fetch `Path` from the flattened `UrlExtract`. I think I will keep `UrlExtract` as is for now, bot with a new name for path.

PR #140, opened 2024-07-26
feat(portal): 139 create server side portal

Tzal3x:
Create a nextjs server that acts as a server-side walrus-sites portal.
It works by combining a middleware for preprocessing the requests, and a root endpoint (defined in`route.ts`) that actually contains the logic as the service worker did. 

The middleware has the following role:
1. To handle requests like `http://<name>.walrus.site/path/to/file/` without throwing an error. Explanation: if you don't define a route on nextjs, the default logic is to throw an error because the `/path/to/file` endpoint does not exist. 
2. To delegate the whole request (including the subnames) to the route.ts in order to fetch the site resources. By default the subnames are removed from the hostname in nextjs. 

In summary, the middleware is just a delegator and an error-preventer of the original url. 

💁‍♂️ **Review tip**: the PR consists of 2 commits. 
I would suggest reviewing commit by commit and not the whole PR altogether.

### DIFF
--- a/portal/server/.prettierrc
+++ b/portal/server/.prettierrc
@@ -1,6 +1,0 @@
-{
-    "trailingComma": "es5",
-    "tabWidth": 4,
-    "semi": false,
-    "singleQuote": true
-}
--- a/portal/server/.prettierrc
+++ b/portal/server/.prettierrc
@@ -1,0 +1,6 @@
+{
+    "trailingComma": "es5",
+    "tabWidth": 4,
+    "semi": false,
+    "singleQuote": true
+}
--- a/portal/server/app/route.ts
+++ b/portal/server/app/route.ts
@@ -1,0 +1,6 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export async function GET(req: Request){
+    return new Response('Soon to be a server portal!', {status: 200, statusText: 'OK'})
+}
--- a/portal/server/app/route.ts
+++ b/portal/server/app/route.ts
@@ -1,6 +1,49 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-export async function GET(req: Request){
-    return new Response('Soon to be a server portal!', {status: 200, statusText: 'OK'})
+import { getDomain, getSubdomainAndPath } from '@lib/domain_parsing'
+import { redirectToAggregatorUrlResponse, redirectToPortalURLResponse } from '@lib/redirects'
+import { getBlobIdLink, getObjectIdLink } from '@lib/links'
+import { resolveAndFetchPage } from '@lib/page_fetching'
+
+export async function GET(req: Request) {
+    const originalUrl = req.headers.get('x-original-url')
+    if (!originalUrl) {
+        throw new Error('No original url found in request headers')
+    }
+    const url = new URL(originalUrl ?? req.url)
+
+    const objectIdPath = getObjectIdLink(url.toString())
+    if (objectIdPath) {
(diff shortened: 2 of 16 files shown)


Tzal3x:
Base skeleton is there. It doesn't work yet, something wrong happens with the url parsing.

Tzal3x:
To check: https://github.com/dcangulo/nextjs-subdomain-example

Tzal3x:
Another article to consider: [nextjs wildcard subdomains](https://www.smashingmagazine.com/2021/11/nextjs-wildcard-subdomains/)

Tzal3x:
Looks like subdomains don't work with safari. 

i.e. http://docs.localhost:3000/ returns this error:
`Safari can't find server. Safari can't find the server docs.localhost`

portal/server/.prettierrc
giac-mysten, line 1: Suggestion: I would prefer this to be integrated with `.editorconfig` . Afaics Prettier does parse the editorconfig and uses it for formatting.

portal/server/app/route.ts
```diff
@@ -0,0 +1,49 @@
...
+import { getDomain, getSubdomainAndPath } from '@lib/domain_parsing'
+import { redirectToAggregatorUrlResponse, redirectToPortalURLResponse } from '@lib/redirects'
+import { getBlobIdLink, getObjectIdLink } from '@lib/links'
+import { resolveAndFetchPage } from '@lib/page_fetching'
+
+export async function GET(req: Request) {
+    const originalUrl = req.headers.get('x-original-url')
+    if (!originalUrl) {
+        throw new Error('No original url found in request headers')
+    }
```
giac-mysten, line 13: Question: why do we need this? do we expect proxies changing the request?

portal/server/custom.d.ts
giac-mysten, line 1: Question: I'm guessing there was no way to import this definition from from `common/`? (probably not a big deal to have this file duplicated)
(edit: apparently there is a way, by exporting this as type definitions in the `package.json`)

portal/server/middleware.ts
giac-mysten, line 1: Question: do we still need the middleware?

giac-mysten (commented):
great stuff @Tzal3x . I made a first pass on the PR. We can address the problems you mention above offline.

portal/server/middleware.ts
Tzal3x: Yes because it needs to delegate the original url. If we don't use it, nextjs removes any subnames, and also it will throw an error if you try to access a path (endpoint) that does not exist like `docs.portal/path/to/something`

portal/server/app/route.ts
Tzal3x: `x-original-url` is a custom header that contains the original url that was requested by the browser. 
I had to use it because there was no other way accessing the subdomains.

portal/server/.prettierrc
Tzal3x: You are right, a lot of things contained in prettier are also used in .editorconfig. 
That will be an easy fix.

portal/server/custom.d.ts
Tzal3x: Cool, I managed to make it work using a [types](https://www.typescriptlang.org/tsconfig/#types) declaration in `server/tsconfig.json`. 

For some reason it doesn't work by defining this in `package.json`, even though we are given that option. 🤷

portal/server/.prettierrc
giac-mysten: did you add the settings to `.editorconfig`? if so, can you commit them?

giac-mysten (approved):
Thanks! Let's merge it as is, and investigate further if we can remove the middleware (#142 )

Tzal3x:
> did you add the settings to .editorconfig? if so, can you commit them?

@giac-mysten I actually juste removed the prettierrc file because it was redundant after I compared them.

Issue #147, opened 2024-08-01
[Bug]: Unable to fetch landing page on server portal


Tzal3x:
### What happened?

When browsing on `http://portal.name/` the walrus sites landing page should be displayed. 

At the moment this is not possible, creating infinite loops on the server. 

### Search terms

x

### Steps to reproduce

1. Deploy the server portal: `cd portal/server && pnpm run dev`
2. Access http://localhost:3000

### Version

- [X] `walrus` and `site-builder` are up to date.

### If applicable, what browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

giac-mysten:
Closed by #148 and #153

PR #148, opened 2024-08-01
fix(portal): fix landing page server portal infinite loop

Tzal3x:
Resolves https://github.com/MystenLabs/walrus-sites/issues/147

## Changes 

Add the functionality so that while browsing the landing page of the server portal fetch the page contents by the worker portal.

## Future work 
This is a temporary solution in order to make the server side portal work. 

As an improvement to this, in a following PR we can consider redirecting to fetch the page from another service on vercel (similar to how the worker portal works).

### DIFF
--- a/portal/common/lib/domain_parsing.ts
+++ b/portal/common/lib/domain_parsing.ts
@@ -62,6 +62,6 @@ function splitUrl(url: URL): UrlExtract {
  * @param path The path to remove the last forward-slash from.
  * @returns The path without the last forward-slash.
  */
-function removeLastSlash(path: string): string {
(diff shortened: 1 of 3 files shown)


Tzal3x:
⚠️ Note: I decided to fetch the contents of the landing page to the server portal instead of redirecting because the server portal will have use a different domain (onwalr.us), and if I redirected then the user would go to walrus.site which would imply  the usage of the worker portal.

giac-mysten (approved):
Thanks for the fix @Tzal3x . I'm not 100% convinced of using the other portal for the landing page, but we can merge this to `develop` to start iterating.

PR #149, opened 2024-08-02
feat(portal): add more unit tests to the portal common library

Tzal3x:


portal/common/lib/aggregator.test.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, test } from "vitest";
+import { aggregatorEndpoint } from "./aggregator";
+
+describe("aggregatorEndpoint", () => {
+    test("blob_id -> URL", () => {
+        const expected = 'https://aggregator-devnet.walrus.space/v1/blob_id';
```
giac-mysten, line 9: Remark: `aggregatorEndpoint` uses the `AGGREGATOR` constant, which may be changed. In that case the test would fail.

Suggestion: We could either (i) have `aggregatorEnpoint` accept the aggregator url as input, or (ii) use the `AGGREGATOR` variable to construct the `expected`  test above. I am slightly in favor of the former.

giac-mysten (approved):
Thanks @Tzal3x !

portal/common/lib/aggregator.test.ts
Tzal3x: Valid point, but this is the point of the test, to fail in case something changes, and update the test when that change happens. If I pass the aggregator constant to the `aggregatorEndpoint` arguments, that would require dynamically constructing the expected value too, which would look like this: 
```typescript
const expected = URL(AGGREGATOR + "/v1/" + encodeURIComponent(blob_id)
```
but the aggregatorEndpoint implementation is already like this 
```typescript
export function aggregatorEndpoint(blob_id: string): URL {
    return new URL(AGGREGATOR + "/v1/" + encodeURIComponent(blob_id));
}
```
Making the test in the end redundant no?

giac-mysten: > Making the test in the end redundant no?

you're right about this

my only fear is that we then provide this code to third parties, and this test will be instantly broken as soon as they change the aggregator endpoint. Maybe no test is better at this point? Up to your judgement :) feel free to merge.

Tzal3x: If that's the case then it would be better removing the test. Then we merge 🙆‍♂️

PR #154, opened 2024-08-07
fix(portal): fetch walrus sites index sw enabled

Tzal3x:
While fetching `walrus.site/index.html` is registering a service worker, which
we don't want in our server portal implementation.

`index-sw-enabled.html` in the same, but without the service worker registration.


giac-mysten (approved):
🔝

PR #159, opened 2024-08-12
Merge develop into main

Tzal3x:
Introduces the server side portal to main


giac-mysten (approved):
Thanks so much for implementing the server-side portal, @Tzal3x ! 🥳 this is very cool to have

PR #161, opened 2024-08-12
refactor: resolve conflicts between develop and main

Tzal3x:
Resolve merge conflicts and introduce the server portal to the main branch.


portal/common/lib/http/http_error_responses.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import template_404 from "@static/404-page.template.html";
+import template_404 from "../../static/404-page.template.html";
```
giac-mysten, line 4: question: I thought we wanted to use `@` imports, why the revert?

portal/server/tsconfig.json
```diff
@@ -26,4 +26,4 @@
     },
     "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
     "exclude": ["node_modules"]
-}
+}
```
giac-mysten, line 29: Question: why remove the newline?

giac-mysten (approved):

portal/common/lib/http/http_error_responses.ts
Tzal3x: We were using `@` imports as aliases for relative imports. 
When we started installing the common lib as a dependency, these aliases stopped working so I reverted them. 
It's fine because relative imports are only used now inside the common lib, where most files are inside the `lib/` dir, with the exception of `static/`.

PR #180, opened 2024-08-22
refactor(portal): Rename index-sw-enabled.html upon publish landing page

Tzal3x:
While publishing the landing page to walrus,
the `index-sw-enabled.html` is renamed to `index.html` to avoid resolving this mapping on
`portal/server/app/route.ts`.


giac-mysten (approved):
Very nice, thanks for the quick change @Tzal3x !

PR #183, opened 2024-08-22
feat(portal): add caching to the service worker

Tzal3x:
Create a caching mechanism on the service worker. 

Caches are expired when:

1. The timestamp is too old
2. The cached sui object version of the `site::Resource` does not match the one that is on-chain. 

~~The `caches` api is only available for "secure contexts" so in order to test this locally one needs to use a tool like `mkcert` for local https certs.~~ 

Localhost is a secure context, so no need to issue a cert to test this. If you still experience issues with the browser's state, use chrome's incognito mode.

### DIFF
--- a/portal/common/lib/resource.ts
+++ b/portal/common/lib/resource.ts
@@ -88,6 +88,7 @@ export async function fetchResource(
     const versionedSiteResource = {
         ...siteResource,
         version: pageData.data?.version,
+        objectId: dynamicFields.data.objectId,
     };
     return versionedSiteResource;
 }
--- a/portal/common/lib/resource.ts
+++ b/portal/common/lib/resource.ts
@@ -3,7 +3,7 @@
 
 import { HttpStatusCodes } from "./http/http_status_codes";
 import { SuiClient, SuiObjectData } from "@mysten/sui/client";
-import { Resource } from "./types";
+import { Resource, VersionedResource } from "./types";
 import { MAX_REDIRECT_DEPTH, RESOURCE_PATH_MOVE_TYPE } from "./constants";
 import { checkRedirect } from "./redirects";
 import { fromB64 } from "@mysten/bcs";
@@ -37,7 +37,7 @@ export async function fetchResource(
     path: string,
     seenResources: Set<string>,
     depth: number = 0,
-): Promise<Resource | HttpStatusCodes> {
+): Promise<VersionedResource | HttpStatusCodes> {
     if (seenResources.has(objectId)) {
         return HttpStatusCodes.LOOP_DETECTED;
     } else if (depth >= MAX_REDIRECT_DEPTH) {
@@ -85,7 +85,11 @@ export async function fetchResource(
     if (!siteResource || !siteResource.blob_id) {
         return HttpStatusCodes.NOT_FOUND;
     }
-    return siteResource;
+    const versionedSiteResource = {
+        ...siteResource,
+        version: pageData.data?.version,
+    };
+    return versionedSiteResource;
 }
 
 /**
--- a/portal/worker/src/caching.ts
+++ b/portal/worker/src/caching.ts
@@ -1,0 +1,100 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { resolveAndFetchPage } from "@lib/page_fetching";
+import { SuiClient, getFullnodeUrl } from "@mysten/sui/client";
+import { NETWORK } from "@lib/constants";
+import { DomainDetails } from "@lib/types";
+
+const CACHE_NAME = "walrus-sites-cache";
+// TODO - move it to .env
+const CACHE_EXPIRATION_TIME = 24 * 60 * 60 * 1000 // 24 hours in milliseconds
+
+/**
+* Respond to the request using the cache API.
+*/
+export default async function respondUsingCache(
+    event: FetchEvent, parsedUrl: DomainDetails, urlString: string
+) {
+    event.respondWith((async () => {
+        if (!('caches' in self)) {
+            // When not being in a secure context, the Cache API is not available.
+            console.warn('Cache API not available');
+            return await resolveAndFetchPage(parsedUrl);
+        }
+
+        const cache = await caches.open(CACHE_NAME);
+        const cachedResponse = await cache.match(urlString);
+        const cacheWasFresh = !(await cleanExpiredCache(cachedResponse, urlString));
+
+        let isCacheSameAsNetwork: boolean;
+        if (cachedResponse && cacheWasFresh) {
+            console.log('Cache hit!')
+            try {
+                isCacheSameAsNetwork = await checkCachedVersionMatchesOnChain(cachedResponse);
+                if (isCacheSameAsNetwork) return cachedResponse;
+            } catch (e) {
+                console.error("Error checking cache version against chain:", e);
+            }
+        }
+        console.log("Cache miss!", urlString);
+        const resolvedPage = await resolveAndFetchPage(parsedUrl);
+
+        cache.put(urlString, resolvedPage.clone());
+        return resolvedPage;
+    })());
+}
+
+/**
+* Removes an entry of the cache, if that entry is expired.
+*
+* The expiration time is set by the `CACHE_EXPIRATION_TIME` constant.
+* If the cached response is older than the expiration time, it's no longer
+* "fresh" and it is removed from the cache.
+*
+* @param urlString the key of the cached entry to check
+* @returns true if the cache entry was removed, false otherwise
+*/
+async function cleanExpiredCache(cachedResponse: Response, urlString: string): Promise<boolean> {
+    const cache = await caches.open(CACHE_NAME);
+    const now = Date.now();
+
+    if (cachedResponse) { // Cache hit!
+        const timestamp = parseInt(cachedResponse.headers.get("x-unix-time-cached") || "0");
+        const hasExpired = now - timestamp > CACHE_EXPIRATION_TIME
+        if (hasExpired) {
+            await cache.delete(urlString);
+            console.log('Removed expired cache entry:', urlString);
+            return true;
+        }
+        console.log('Cache entry is still fresh:', urlString)
+    }
+    return false;
+}
+
+/**
+* Check if the cached version of the Resource object matches the current on-chain version.
+*
+* @param cachedResponse the response to check the version of
+* @returns true if the cached version matches the current version of the Resource object
+*/
+async function checkCachedVersionMatchesOnChain(cachedResponse: Response): Promise<boolean> {
+    if (!cachedResponse){
+        throw new Error("Cached response is null!");
+    }
+    const rpcUrl = getFullnodeUrl(NETWORK);
+    const client = new SuiClient({ url: rpcUrl });
+    const cachedVersion = cachedResponse.headers.get("x-resource-sui-object-version")
+    const objectId = cachedResponse.headers.get("x-resource-sui-object-id");
+    if (!cachedVersion || !objectId) {
+        throw new Error("Cached response does not have the required headers");
+    }
+    const resourceObject = await client.getObject({id: objectId});
+    if (!resourceObject.data) {
+        throw new Error("Could not retrieve Resource object.");
+    }
+    console.log("Cached version: ", cachedVersion)
+    console.log("Current version: ", resourceObject.data?.version)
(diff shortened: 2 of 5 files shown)


Tzal3x:
Getting some `Cache miss!` even though I put those URLs to the cache!

Probably an inconsistency between saved cache keys, and search keys used in `cache.match`.  💭

portal/common/lib/resource.ts
giac-mysten, line 1: Nit: you may need to rebase on main :)

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -50,8 +54,40 @@ self.addEventListener("fetch", async (event) => {
     console.log("Parsed URL: ", parsedUrl);
 
     if (requestDomain == portalDomain && parsedUrl && parsedUrl.subdomain) {
-        console.log("fetching from the service worker");
-        event.respondWith(resolveAndFetchPage(parsedUrl));
+        event.respondWith((async () => {
+            // Clean the cache of expired entries
+            await cleanExpiredCache();
```
giac-mysten, line 59: Remark: Here, we are incurring the cost of checking the whole list every time we fetch a resource. This is probably unnecessary.

Suggestion: Instead of checking the whole list
- check if there is a match in the cache for the requested site;
  - if there is a match, check its expiration:
    - if it's expired, remove the entry from the cache
    - if it's not expired, use the entry
  - if there is no match, fetch directly

this way, we still have the property that the entries we fetch from the cache are "fresh", but for an O(1) cost instead of O(n).

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -50,8 +54,40 @@ self.addEventListener("fetch", async (event) => {
...
+                console.warn('Cache API not available');
+                return await resolveAndFetchPage(parsedUrl);
+            }
+
+            const cache = await caches.open(cacheName);
+            const cachedResponse = await cache.match(urlString);
+            let isCacheSameAsNetwork: boolean;
+            try {
+                if (cachedResponse) {
+                    isCacheSameAsNetwork = await checkCachedVersionMatchesOnChain(cachedResponse)
```
giac-mysten, line 71: Nit
```suggestion
                    isCacheSameAsNetwork = await checkCachedVersionMatchesOnChain(cachedResponse);
```

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -50,8 +54,40 @@ self.addEventListener("fetch", async (event) => {
...
+                }
+            } catch (e) {
+                console.error("Error checking cache version against chain:", e);
+            }
+            if (cachedResponse && isCacheSameAsNetwork) {
```
giac-mysten, line 76: Question: i think `isCacheSameAsNetwork` may be uninitialized at this point?

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -50,8 +54,40 @@ self.addEventListener("fetch", async (event) => {
...
+                console.log("Cache hit!", urlString);
+                return cachedResponse;
+            } else {
+                console.log("Cache miss!", urlString);
+                const resolvedPage = await resolveAndFetchPage(parsedUrl);
+
+                // The urlString (i.e. the cache key) will also contain the timestamp
+                // to ensure that the cache is not stale.
+                // The delimiter used is an illegal URI character
+                // to avoid any conflicts with the actual URL.
```
giac-mysten, line 86: Question: where is this timestamp set in the url? Is this an outdated comment?

portal/worker/src/walrus-sites-sw.ts
giac-mysten: (we may still want to periodically trigger a full cache clean)

portal/worker/src/walrus-sites-sw.ts
giac-mysten, line 71: Note for the future: we may want to optimize this (as discussed) and not always check if there is an updated version -- which costs another RTT to the full node.

The other problem of caching based on the object version is that the object version may change, without the object itself changing at all, triggering a re-fetch.

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -50,8 +54,40 @@ self.addEventListener("fetch", async (event) => {
...
+                cache.put(urlString, resolvedPage.clone());
+                return resolvedPage;
+            }
```
giac-mysten, line 89: Suggestion: move this to a separate function

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -5,14 +5,18 @@ import { getDomain, getSubdomainAndPath } from "@lib/domain_parsing";
 import { redirectToAggregatorUrlResponse, redirectToPortalURLResponse } from "@lib/redirects";
 import { getBlobIdLink, getObjectIdLink } from "@lib/links";
 import { resolveAndFetchPage } from "@lib/page_fetching";
+import { SuiClient, getFullnodeUrl } from "@mysten/sui/client";
+import { NETWORK } from "@lib/constants";
+
+const cacheName = "walrus-sites-cache";
```
giac-mysten, line 11:
```suggestion
const CACHE_NAME = "walrus-sites-cache";
```

giac-mysten (changes requested):
Nice stuff! I left a few suggestions. 
Going forward we can discuss how to introduce caching also for aggregator responses!

giac-mysten:
First informal benchmark, using the docs site.
We compare here loading and refreshing a site three times (starting from a fresh incognito browser) with and without cache implemented in the service worker.

The cache clearly speeds up loading content after the first time the page is loaded.

| Cache? | Page load | Finish (s) | DOMContentLoaded (ms) | load (ms) |
|--------|-----------|------------|-----------------------|-----------|
|        | 1         | 2.65       | 519                   | 991       |
| No     | 2         | 1.55       | 568                   | 838       |
|        | 3         | 1.53       | 484                   | 663       |
|--------|-----------|------------|-----------------------|-----------|
|        | 1         | 2.03       | 561                   | 745       |
| Yes    | 2         | 1.43       | 175                   | 221       |
|        | 3         | 1.42       | 167                   | 224       |

portal/worker/src/walrus-sites-sw.ts
Tzal3x: It's true that there can be a case where the `isCacheSameAsNetwork` can be uninitialized, but this is not an issue in typescript, because uninitialized == `undefined`.

e.g. 
```typescript
let x: boolean;
x ? true : false; // Will return false, because boolean operations with `undefined` result to `false` in js/ts
```

portal/worker/src/walrus-sites-sw.ts
Tzal3x: Good catch, that's a remnant of the first implementation. I will remove it ✍️

portal/worker/src/walrus-sites-sw.ts
Tzal3x: Created an issue for this: https://github.com/MystenLabs/walrus-sites/issues/194

portal/worker/src/walrus-sites-sw.ts
giac-mysten: omg javascript 🙄

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -50,8 +48,7 @@ self.addEventListener("fetch", async (event) => {
     console.log("Parsed URL: ", parsedUrl);
 
     if (requestDomain == portalDomain && parsedUrl && parsedUrl.subdomain) {
-        console.log("fetching from the service worker");
-        event.respondWith(resolveAndFetchPage(parsedUrl));
+        await respondUsingCache(event, parsedUrl, urlString);
```
giac-mysten, line 51: Suggestion: for consistency and readability, I would prefer something along these lines:

```suggestion
    if (requestDomain == portalDomain && parsedUrl && parsedUrl.subdomain) {
        if (!("caches" in self)) {
            // When not being in a secure context, the Cache API is not available.
            console.warn("Cache API not available");
            return await resolveAndFetchPage(parsedUrl);
        }
        return await resolveWithCache(parsedUrl);
    }
```

this way it is clear when the cache is used, and we can avoid passing the event.
Further, it may be nice to add the full `urlString` to the `DomainDetails` struct, so that the function signature is similar?

portal/worker/src/caching.ts
```diff
@@ -0,0 +1,100 @@
...
+const CACHE_NAME = "walrus-sites-cache";
+// TODO - move it to .env
+const CACHE_EXPIRATION_TIME = 24 * 60 * 60 * 1000 // 24 hours in milliseconds
+
+/**
+* Respond to the request using the cache API.
+*/
+export default async function respondUsingCache(
+    event: FetchEvent, parsedUrl: DomainDetails, urlString: string
+) {
```
giac-mysten, line 18: Suggestions: with the changes above, the signature can be simplified to
```suggestion
export default async function resolveWithCache(parsedUrl: DomainDetails): Promise<Response> {
```
and the internal `event.respondWith` can be removed

portal/worker/src/caching.ts
```diff
@@ -0,0 +1,100 @@
...
+/**
+* Check if the cached version of the Resource object matches the current on-chain version.
+*
+* @param cachedResponse the response to check the version of
+* @returns true if the cached version matches the current version of the Resource object
+*/
+async function checkCachedVersionMatchesOnChain(cachedResponse: Response): Promise<boolean> {
+    if (!cachedResponse){
+        throw new Error("Cached response is null!");
+    }
```
giac-mysten, line 84: Question: since `cachedResponse: Response` , this should not be possible, right?

giac-mysten (approved):
Thanks for the changes, @Tzal3x ! Just a few additional suggestions from my side, up to you what you want to take.

portal/worker/src/caching.ts
Tzal3x: Fair 💯

portal/worker/src/caching.ts
Tzal3x: Just a sanity check, I would prefer to keep it this way, in case the type checking is somehow bypassed (e.g. calling it from the `dist/` JS code etc).

portal/worker/src/walrus-sites-sw.ts
Tzal3x: I would avoid touching for now the `DomainDetails` type to not affect any server portal functionality.

portal/worker/src/caching.ts
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { resolveAndFetchPage } from "@lib/page_fetching";
+import { SuiClient, getFullnodeUrl } from "@mysten/sui/client";
+import { NETWORK } from "@lib/constants";
+import { DomainDetails } from "@lib/types";
+
+const CACHE_NAME = "walrus-sites-cache";
+// TODO - move it to .env
```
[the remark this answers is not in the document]
Tzal3x: I thought this over, and since the service worker is a one-time install, we should not be changing the `.env` arbitrarily like that without this change showing up somewhere. 

So let's keep it inside the code so that it passes through a PR whenever we need to change something like this.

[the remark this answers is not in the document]
giac-mysten: makes sense. Then remove the todo :)

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -50,9 +49,15 @@ self.addEventListener("fetch", async (event) => {
...
-        return;
+        if (!("caches" in self)) {
+            // When not being in a secure context, the Cache API is not available.
+            console.warn("Cache API not available");
+            const response = resolveAndFetchPage(parsedUrl)
+            event.respondWith(response);
+            return
+        }
+        const response = resolveWithCache(parsedUrl, urlString)
+        return event.respondWith(response);
```
giac-mysten, line 60: Nit: a bit of dedup
```suggestion
        let response: Promise<Response>;
        if (!("caches" in self)) {
            // When not being in a secure context, the Cache API is not available.
            console.warn("Cache API not available");
            response = resolveAndFetchPage(parsedUrl);
        } else {
            response = resolveWithCache(parsedUrl, urlString);
        }
        event.respondWith(response);
        return;

```

giac-mysten: or even `return   event.respondWith(response);` should be ok?

PR #184, opened 2024-08-23
refactor(portal): resolve the objectId to a separate function

Tzal3x:
This simplifies the code of the `resolveAndFetchPage()`.


portal/common/lib/page_fetching.ts
```diff
@@ -17,7 +17,18 @@ import { aggregatorEndpoint } from "./aggregator";
 export async function resolveAndFetchPage(parsedUrl: DomainDetails): Promise<Response> {
     const rpcUrl = getFullnodeUrl(NETWORK);
     const client = new SuiClient({ url: rpcUrl });
+    const objectId = await resolveObjectId(parsedUrl, client);
+    if (typeof objectId == "string") {
```
giac-mysten, line 21: Question: I think we lose the `fullNodeFail()` case, right?

Tzal3x: Good catch @giac-mysten ! I just patched it

giac-mysten (approved):
Thanks, LGTM!

PR #187, opened 2024-08-26
Introduce LANDING_PAGE_BLOB_ID and install @types/node

Tzal3x:
Use an environment variable (`LANDING_PAGE_BLOB_ID`) for making it easy to specify the walrus sites blob id of the landing page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ env/
 .next/
 *.log
 temp-landing-page/
+.env*.local
--- a/portal/server/app/route.ts
+++ b/portal/server/app/route.ts
@@ -37,7 +37,7 @@ export async function GET(req: Request) {
     const atBaseUrl = portalDomain == url.host.split(':')[0]
     if (atBaseUrl) {
         console.log('Serving the landing page from walrus...')
-        const blobId = '60y9fj0iyk5nt1pu3uebk6ppivknyw04wvx9rbqwqpai5jjwu6'
+        const blobId = process.env.LANDING_PAGE_BLOB_ID!;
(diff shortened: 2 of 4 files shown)


portal/server/app/route.ts
```diff
@@ -37,7 +37,7 @@ export async function GET(req: Request) {
     const atBaseUrl = portalDomain == url.host.split(':')[0]
     if (atBaseUrl) {
         console.log('Serving the landing page from walrus...')
-        const blobId = '60y9fj0iyk5nt1pu3uebk6ppivknyw04wvx9rbqwqpai5jjwu6'
+        const blobId = process.env.LANDING_PAGE_BLOB_ID!;
```
giac-mysten, line 40: Important: this is not a blob id. It's the Base36-encoded version of the object ID of the walrus site
```suggestion
        const blobId = process.env.LANDING_PAGE_OID_B36!;
```

.gitignore
```diff
@@ -14,3 +14,4 @@ env/
 .next/
 *.log
 temp-landing-page/
+.env*.local
```
giac-mysten, line 17: Question: why this pattern for the environment variable? is this some standard web frameworks use?

giac-mysten (approved):
Thanks @Tzal3x ! this works :)

.gitignore
Tzal3x: Correct! Whenever you run `vercel env pull` the name of the environment variables file is `.env.local`. 
There can also be variations to this name depending on our vercel setup.
So to git-ignore it, you must much use such patterns.

PR #192, opened 2024-08-27
feat(service-worker): Redirect to blob.store if SW not able to install

Tzal3x:
Some browsers (like in-app browsers) don't support service workers.

Instead of throwing an error to the user, we
redirect to `blob.store`, where the resolution of the site resources happens on the server side.

### DIFF
--- a/portal/common/static/walrus-sites-portal-register-sw.js
+++ b/portal/common/static/walrus-sites-portal-register-sw.js
@@ -24,20 +24,14 @@ function main() {
             })
             .catch(handleError);
     } else {
-        displayErrorMessage(swNotSupportedNode());
+        const currentUrl = new URL(window.location.href);
+        console.warn("This browser does not yet support Walrus Sites 💔, redirecting to blob.store");
+        window.location.href = currentUrl.toString().replace('walrus.site', 'blob.store');
     }
 }
 
 function handleError(error) {
     displayErrorMessage(swNotLoadingNode());
-}
-
-function swNotSupportedNode() {
-    return titleSubtitleNode(
-        "This browser does not yet support Walrus Sites 💔",
-        'Please try using a different browser, such as Chrome, Firefox (not in "Private mode"), \
-        or Safari.'
-    );
 }
 
 function swNotLoadingNode() {
--- a/portal/common/lib/constants.ts
+++ b/portal/common/lib/constants.ts
@@ -10,5 +10,7 @@ export const SITE_NAMES: { [key: string]: string } = {
     // e.g.,
     // landing: "0x1234..."
 };
+// The default portal to redirect to if the browser does not support service workers.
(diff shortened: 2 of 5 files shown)


portal/common/static/walrus-sites-portal-register-sw.js
```diff
@@ -24,22 +24,16 @@ function main() {
             })
             .catch(handleError);
     } else {
-        displayErrorMessage(swNotSupportedNode());
+        const currentUrl = new URL(window.location.href);
+        console.warn("This browser does not yet support Walrus Sites 💔, redirecting to blob.store");
+        window.location.href = currentUrl.toString().replace('walrus.site', 'blob.store');
```
giac-mysten, line 29: Important: using `walrus.site` breaks the portal when hosting on `localhost`
Suggestion: use `getPortalUrl`?

Other suggestion: make `blob.store` a constant, maybe under `constants.ts`, so that it is easy to customize for partners.

giac-mysten (approved):
Thanks for the quick fix, really cool  that we can do this so easily :) (cc @Jordan-Mysten )

portal/common/static/walrus-sites-portal-register-sw.js
Jordan-Mysten: I think a better way to do this would be:

```ts
window.location.href = new URL(
  `${currentUrl.pathname}${currentUrl.search}${currentUrl.hash}`,
  'https://blob.store',
);
```

this is more correct than basic string replacement.

Tzal3x: > I think a better way to do this would be:
> 
> ```ts
> window.location.href = new URL(
>   `${currentUrl.pathname}${currentUrl.search}${currentUrl.hash}`,
>   'https://blob.store',
> );
> ```
> 
> this is more correct than basic string replacement.

The problem with this solution is that the subdomain is not parsed. If we need to access docs.walrus.site, then the base of the url should be docs.blob.store, not just blob.store.

Tzal3x: > Important: using `walrus.site` breaks the portal when hosting on `localhost` Suggestion: use `getPortalUrl`?

Importing `getPortalUrl` is tricky in this script because it's a js file, and `getPortalUrl` is part of a ts library. We would need to bundle (transpile to js) the whole common/lib library in order to import something from there. 

> Other suggestion: make `blob.store` a constant, maybe under `constants.ts`, so that it is easy to customize for partners.

Yes that could be an easy addition! 👍

Tzal3x: So I combined both of the recommendations and moved the `blob.store` to a `FALLBACK_PORTAL` constant. 

Let me know what you think 🙆‍♂️

PR #202, opened 2024-09-09
Fix missing newline in the 404 page

Tzal3x:
The message shown would be like the
following:

"You have ... internet!Please turn back!"
In order to add the newline, I would have
break this sentence into two, because
in HTML cannot be added simply
including the "\n" character.

### DIFF
--- a/portal/common/lib/http/http_error_responses.ts
+++ b/portal/common/lib/http/http_error_responses.ts
@@ -5,23 +5,27 @@ import template_404 from "../../static/404-page.template.html";
 
 export function siteNotFound(): Response {
     return Response404(
-        "You have reached the end of the internet, please turn back!" +
+        "You have reached the end of the internet, please turn back!",
         "Invalid URL: The object ID is not a valid Walrus Site."
     );
 }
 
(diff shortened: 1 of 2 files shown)


portal/common/lib/http/http_error_responses.ts
```diff
@@ -5,23 +5,27 @@ import template_404 from "../../static/404-page.template.html";
...
+        "You have reached the end of the internet, please turn back!",
         "Invalid URL: The object ID is not a valid Walrus Site."
     );
 }
 
 export function noObjectIdFound(): Response {
-    return Response404("You have reached the end of the internet, please turn back!" +
-        "Invalid URL: No object ID could be found.");
+    return Response404(
+        "You have reached the end of the internet, please turn back!",
```
giac-mysten, line 15: Nit: use const for this, since it is used twice

Nit: 
```suggestion
        "You have reached the end of the Internet, please turn back!",
```

giac-mysten (approved):
Thanks @Tzal3x ! this works, just two very minor nits.

PR #203, opened 2024-09-09
fix(portal): extract portal link

### LINKED ISSUE #190: The portal not found page links to walrus.site instead of blob.store
It should instead link to the root url of the portal.

image

Tzal3x:
Closes #190 

Page 404 would provide a link for
walrus.site, even when the server side
portal is used (blob.store).

## Changes
Created a new function that extracts the 
appropriate redirection
link based on the current URL in order to
point to the correct portal, and uses it
to the `<a href=...>` tag.

i.e. if the user typed `<name>.walrus.site`
he should be redirected to `walrus.site` when
clicking the hyperlink.
Likewise, if the user typed `<name>.blob.store`, 
he should be redirected to `blob.store`.

### DIFF
--- a/portal/common/static/404-page.template.html
+++ b/portal/common/static/404-page.template.html
@@ -12,19 +12,18 @@
     </head>
     <script>
     function getPortalName() {
-        const url = window.location.href;
-        if (url.includes("walrus.site")) {
-            console.log("URL contains 'walrus.site'");
-            return ["https://walrus.site/", "Open walrus.site"];
+        const url = new URL(window.location.href);
+        const domainParts = url.host.split('.');
+        const portalName = domainParts.slice(-2).join('.');
+        if (portalName.includes('localhost')) {
+            return 'http://localhost' + `:${url.port}`
         }
-        return ["https://blob.store/", "Open blob.store"];
+        return `https://${portalName}`
     };
 
     window.onload = function() {
         const portalLink = document.getElementById('portalLink');
-        const [link, text] = getPortalName();
-        portalLink.href = link;
-        portalLink.textContent = text;
+        portalLink.href = getPortalName();
     };
     </script>
     <style>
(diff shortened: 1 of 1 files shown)


portal/common/static/404-page.template.html
```diff
@@ -10,6 +10,23 @@
...
     </head>
+    <script>
+    function getPortalName() {
+        const url = window.location.href;
+        if (url.includes("walrus.site")) {
+            console.log("URL contains 'walrus.site'");
+            return ["https://walrus.site/", "Open walrus.site"];
+        }
+        return ["https://blob.store/", "Open blob.store"];
+    };
```
giac-mysten, line 21: Question: This will only work with `walrus.site` or `blob.store`. What if we wanted to support other (partner) portals? Can we use something like `getPortalUrl` in `common/lib/redirects.ts` (with the appropriate modifications ofc)

Tzal3x: Ah, now I get it, that makes sense. Alright I will make it this way 💯

Tzal3x: Looking more into it, I can't transfer the `getPortalUrl` from `redirects.ts` because it uses the `getDomain` function from `domain_parsing.ts`, which consequently uses the [parse-domain lib](https://www.npmjs.com/package/parse-domain).

To extract properly the portal domain, would require using our `getPortalUrl` because it splits reliably the suins/blobIDs from the portal.

This cannot be done with a few changes. On the other hand there is always the option to go with the old approach of exctracting the portal domain as the last 2 elements of the array that results with `<hostname>.split(".")`

giac-mysten: >On the other hand there is always the option to go with the old approach of extracting the portal domain as the last 2 elements of the array that results with <hostname>.split(".")

I'd go with this?

Tzal3x: Done ✔️

giac-mysten (approved):
Amazing, thanks for the changes!

PR #204, opened 2024-09-10
feat(portal): Bench flatland - fetchPage & fetchResource

Tzal3x:
Bench flatland with fetchPage and fetchResource.

The addition was simple: use the already existing bench files for fetchPage and fetchResource, 
create some abstractions to keep it DRY, and include the flatland WS objectId.

### DIFF
--- a/portal/common/lib/page_fetching.bench.ts
+++ b/portal/common/lib/page_fetching.bench.ts
@@ -5,17 +5,12 @@ import { bench, describe, expect } from 'vitest';
 import { fetchPage } from './page_fetching';
 import { SuiClient, getFullnodeUrl } from '@mysten/sui/client'
 import { NETWORK } from './constants';
+import { SITES_USED_FOR_BENCHING } from './constants';
 
-const LANDING_PAGE_OID = '0x5fa99da7c4af9e2e2d0fb4503b058b9181693e463998c87c40be78fa2a1ca271';
-const FLATLAND_OID = '0x049b6d3f34789904efcc20254400b7dca5548ee35cd7b5b145a211f85b2532fa';
-const ws_pages = [
-    [LANDING_PAGE_OID, "landing page"],
-    [FLATLAND_OID, "flatland"]
-]
 const rpcUrl = getFullnodeUrl(NETWORK);
 const client = new SuiClient({ url: rpcUrl });
 describe('Page fetching', () => {
-    ws_pages.forEach(([objectId, siteName]) => {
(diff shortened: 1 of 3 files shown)


portal/common/lib/resource.bench.ts
```diff
@@ -5,15 +5,22 @@ import { bench, describe, expect } from 'vitest';
...
-        const resourcePath = '/index.html';
-        const resp = await fetchResource(client, objectId, resourcePath, new Set());
-        expect(isVersionedResource(resp)).toBeTruthy();
-    });
+const LANDING_PAGE_OID = '0x5fa99da7c4af9e2e2d0fb4503b058b9181693e463998c87c40be78fa2a1ca271';
+const FLATLAND_OID = '0x049b6d3f34789904efcc20254400b7dca5548ee35cd7b5b145a211f85b2532fa';
+const ws_pages = [
+    [LANDING_PAGE_OID, "landing page"],
+    [FLATLAND_OID, "flatland"]
+]
```
giac-mysten, line 15: Suggestion: Can we deduplicate these with the other bench?

portal/common/lib/page_fetching.bench.ts
```diff
@@ -6,13 +6,20 @@ import { fetchPage } from './page_fetching';
 import { SuiClient, getFullnodeUrl } from '@mysten/sui/client'
 import { NETWORK } from './constants';
 
+const LANDING_PAGE_OID = '0x5fa99da7c4af9e2e2d0fb4503b058b9181693e463998c87c40be78fa2a1ca271';
+const FLATLAND_OID = '0x049b6d3f34789904efcc20254400b7dca5548ee35cd7b5b145a211f85b2532fa';
```
giac-mysten, line 10: Suggestion: this is the object ID of the flatland mint page, which is an interesting thing to bench. However, we had discussed to bench also one of the flatlander sites, so that we bench redirects as well. Can you add one of those here? (mint a new one and copy the object id?)

giac-mysten (commented):
Thanks @Tzal3x ! I have a couple of suggestions for this one :)

In general: how do you feel about benching the object ID resolution as well?

portal/common/lib/page_fetching.bench.ts
Tzal3x: You're right, I missed this one! I will add it asap 👍

giac-mysten (approved):
Thanks @Tzal3x !

PR #206, opened 2024-09-11
refactor: Remove redundant pageData fetch

Tzal3x:
Inside `fetchResource` when a redirectId is found,
 there was a `client.getObject` call to fetch the
`redirectPage.data` and check if the page data exist.

This is redundant, as this is already checked
when fetching the `pageData` a few lines bellow,
so we can emit the aforementioned check, and
directly recur calling `fetchResource`.

Bonus: 
- remove redundant `else` clause
- update fetchResource tests in order to pass (it's no longer
needed to mock all those getObject calls, which is also an indicator
of the performance improvement here).

### DIFF
--- a/portal/common/lib/resource.test.ts
+++ b/portal/common/lib/resource.test.ts
@@ -243,18 +243,6 @@ describe('fetchResource', () => {
             },
         });
 
-        getObject.mockResolvedValueOnce({
-            data: {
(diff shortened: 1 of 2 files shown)


giac-mysten (approved):
Excellent, thanks @Tzal3x ! I wonder why that check was there 🤔

PR #207, opened 2024-09-12
fix(codspeed): use main instead of master

Tzal3x:
The fact that the branch that codspeed monitors was set to **master**, resulted in the performance graphs not showing to the dashboard: https://codspeed.io/MystenLabs/walrus-sites/benchmarks

With that quick fix this should be fine now.


giac-mysten (approved):
Thanks @Tzal3x !

PR #208, opened 2024-09-12
feat(portal): Improve fetchResource performance

Tzal3x:
Performance improvement! 🏁

`Promise.all` is as slow as it is the slowest of
the Promises it contains.

So far we have been combining `checkRedirect` and `dynamicFieldObject` fetching
in the same `Promise.all` clause. But `checkRedirect` is quite slow, and it's a case
that is not that frequent (currently only flatland makes use of it).

Given the above, we could start by fetching the dynamicFieldObject first. 

If an object has `dynamicFields` of `RESOURCE_PATH_MOVE_TYPE` it means it's a resource,
which also means that **there won't be any redirect**.

On the other hand, if the "dynamicField fetching" flow fails, then an error will be thrown in order to be caught by the `checkRedirect` and check if it's really a "redirect-able" object.

**Bonus changes**
+ refactored the fetchResource a bit
+ updated the fetchResource tests

### DIFF
--- a/portal/common/lib/resource.test.ts
+++ b/portal/common/lib/resource.test.ts
@@ -65,8 +65,6 @@ describe('fetchResource', () => {
         });
 
     test('should fetch resource without redirect', async () => {
-        // Mock no redirect
-        (checkRedirect as any).mockResolvedValueOnce(null);
         // Mock dynamic field response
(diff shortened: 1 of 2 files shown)


giac-mysten (approved):
Great improvement, thanks @Tzal3x  for teaching me how `Promise`s work in js 😅

PR #212, opened 2024-09-16
feat(portal): Include blob_hash to Resource

Tzal3x:
- Add a `blob_hash` field to the site::Resource in the smart contract
- On the site-builder, calculate the blob_hash (using blake2b256) and include it to the Resource while publishing the site
- On the portal side, check that the hash of the response from the aggregator matches the blob_hash

⚠️ Every walrus site needs to be re-deployed since the walrus site smart contract is changed. 

Notes: 
- Codspeed continues to return unstable results - still need to mock network requests. 
- On vercel it seems that there are too many deployments - still containing the old names of the services too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,18 +15,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


Tzal3x:
Benchmark tests fail because I need to redeploy the contract for flatland using the new package Id of walrus sites.

Tzal3x:
Updated the nft-viz site here: https://github.com/MystenLabs/example-walrus-sites/pull/8

Tzal3x:
Closed this PR and creating a new one to try out if it's possible to reset vercel deployements.

PR #219, opened 2024-09-20
feat(portal): add blob hash to resource fields

Tzal3x:
- Add a `blob_hash` field to the site::Resource in the smart contract
- On the site-builder, calculate the blob_hash (using blake2b256) and include it to the Resource while publishing the site
- On the portal side, check that the hash of the response from the aggregator matches the blob_hash

⚠️ Every walrus site needs to be re-deployed since the walrus site smart contract is changed.

Notes:
- Codspeed continues to return unstable results - still need to mock network requests.


portal/common/lib/bcs_data_parsing.ts
```diff
@@ -2,20 +2,27 @@
...
     output: (id) => base64UrlSafeEncode(bcs.u256().serialize(id).toBytes()),
 });
 
+// Different than BLOB_ID, because we don't want this to be URL-safe;
+// otherwise, it will mess up with the checksum results.
+const U256 = bcs.u256().transform({
+    input: (id: string) => id,
+    output: (id) => toB64(bcs.u256().serialize(id).toBytes()),
+});
+
```
giac-mysten, line 25: Question: why B64? usually hashes are displayed in hex?

giac-mysten: Suggestion: keep a fixed-size byte array instead of a string. This would allow (1) not having the conversion to hex here, and (2) not having to covert the `sha256` hash to hex later on, simplifying the comparison.

portal/common/lib/http/http_error_responses.ts
```diff
@@ -34,3 +36,17 @@ function Response404(message: String, secondaryMessage?: String): Response {
         },
     });
 }
+
+/**
+* Returns the html page that displays an alert to the user
+* regarding a mismatch between the aggregator response and
+* the blob hash (checksum).
+*/
+export function generateChecksumErrorResponse(): Response {
```
giac-mysten, line 45: Nit: `Checksum` as a term is weaker than `hash`. Should we just say that the hash does not match? (and rename the functions?)

portal/common/lib/http/http_status_codes.ts
```diff
@@ -5,5 +5,6 @@
 export enum HttpStatusCodes {
     TOO_MANY_REDIRECTS = 310,
     NOT_FOUND = 404,
+    UNPROCESSABLE_ENTITY = 422,
```
giac-mysten, line 8: Nit: 422 is unprocessable content https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

```suggestion
    UNPROCESSABLE_CONTENT = 422,
```

portal/common/lib/page_fetching.ts
```diff
@@ -78,6 +83,19 @@ export async function fetchPage(
 
     // Deserialize the bcs encoded body and decompress.
     const body = await contents.arrayBuffer();
+    // Verify the integrity of the aggregator response by hashing
+    // the response contents.
+    const h10b = sha256(
+        new Uint8Array(body),
+    );
```
giac-mysten, line 90: Remark: see above comment about keeping the hash as a byte array.

portal/common/lib/page_fetching.ts
```diff
@@ -78,6 +83,19 @@ export async function fetchPage(
 
     // Deserialize the bcs encoded body and decompress.
```
giac-mysten, line 84: Nit: this comment is outdated. Decompression happens later.

portal/common/lib/page_fetching.ts
```diff
@@ -78,6 +83,19 @@ export async function fetchPage(
...
+    );
+    if (result.blob_hash != toB64(h10b)) {
+        console.warn(
+            '[!] checksum mismatch [!] for:', result.path, '.',
+            `blob hash: ${result.blob_hash} | aggr. hash: ${toB64(h10b)}`
+        )
+        return generateChecksumErrorResponse()
+    }
+
     const decompressed = await decompressData(new Uint8Array(body), result.content_encoding);
```
giac-mysten, line 99: Important question: here you check the hash before decompression. Is this the intended way. Note that this choice will also influence the site builder.

advantages: hashing a compressed string can be faster.
disadvantages: this hash does not match the SRI hash anymore.

portal/common/package.json
```diff
@@ -2,6 +2,7 @@
 		"dependencies": {
 				"@mysten/bcs": "^1.0.2",
 				"@mysten/sui": "^1.1.2",
+				"@noble/hashes": "^1.5.0",
```
giac-mysten, line 5: Question: could we use the [web cryptography api](https://www.w3.org/TR/WebCryptoAPI/#dfn-SubtleCrypto-method-digest) standard instead?

site-builder/src/publish.rs
```diff
@@ -173,7 +173,7 @@ pub async fn edit_site(
         "Parsing the directory {} and locally computing blob IDs",
         directory.to_string_lossy()
     ));
-    resource_manager.read_dir(directory, content_encoding)?;
+    resource_manager.read_dir(directory, content_encoding)?; // [!]
```
giac-mysten, line 176: Nit: Leftover comment?

site-builder/src/site/resource.rs
```diff
@@ -373,6 +387,7 @@ impl ResourceManager {
             content_type,
             *content_encoding,
             output.blob_id,
+            U256::from_le_bytes(&blob_hash),
```
giac-mysten, line 390: Question: any specific reason behind the use of `le` bytes?

site-builder/Cargo.toml
```diff
@@ -11,8 +11,10 @@ base64 = "0.22.1"
 bcs = "0.1.6"
 clap = { version = "4.5.7", features = ["derive"] }
 crossterm = "0.27.0"
+fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto" }
 flate2 = "1.0.30"
 futures = "0.3.30"
+hex = "0.4.3"
```
giac-mysten, line 17: Question: Is this use anywhere? Otherwise, remove.

giac-mysten (commented):
Thanks @Tzal3x ! A first round of comments :)

portal/common/package.json
Tzal3x: Correct, we could do that, found an example on MDN docs too: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#supported_algorithms 

I will replace it.

site-builder/src/site/resource.rs
Tzal3x: That was the only way I could find to cast the byte array to U256.

site-builder/Cargo.toml
Tzal3x: Good catch! That was a remnant from the initial code that was using hex encodings. I will remove it asap 👍

portal/common/lib/bcs_data_parsing.ts
Tzal3x: `toBytes` returns a `Uint8Array`. Uint8Array is a subclass of the hidden [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray) class.

There is no standard way of comparing two typed arrays, so we would have to compare them like so: 
```ts
function compareUint8Arrays(arr1, arr2) {
  if (arr1.length !== arr2.length) return false;
  return arr1.every((value, index) => value === arr2[index]);
}
```

or like this: 

```typescript
function compareUint8Arrays(arr1, arr2) {
  return Buffer.compare(Buffer.from(arr1), Buffer.from(arr2)) === 0;
}
```

On the other hand, with the existing implementation we can just compare 2 Base64 strings, which to me looks simpler.

Do we still need to use the Uint8Array approach? If so what would be the benefit?

portal/common/lib/page_fetching.ts
Tzal3x: Hmm that's true, I think I would go with checking the hash _after_ decompression.

site-builder/src/site/resource.rs
giac-mysten: ah yes, now I recall that BCS encodes as LE bytes.

portal/common/lib/bcs_data_parsing.ts
giac-mysten: mmm I see. this is a very annoying language quirk?
Ok with strings then. But can we use Hex instead for the hash?

portal/common/lib/crypto.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+const { subtle } = globalThis.crypto;
```
giac-mysten, line 4: 😂 I love that they called the crypto library `subtle`.

giac-mysten: Question: this works with node/server side js as well, right?

portal/common/lib/crypto.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Calculates SHA-256 hash of input message.
+ * @param message ArrayBuffer to hash
+ * @returns Promise<Uint8Array> Resulting hash as Uint8Array
+ */
+export async function sha256(message: ArrayBuffer) {
```
giac-mysten, line 11: Nit: unsure about this, but shouldn't the return type be specified?
```suggestion
export async function sha256(message: ArrayBuffer): Promise<Uint8Array> {
```

portal/common/lib/bcs_data_parsing.ts
```diff
@@ -2,20 +2,27 @@
...
-// Blob IDs are represented on chain as u256, but serialized in URLs as URL-safe Base64.
+// Blob IDs & hashes are represented on chain as u256, but serialized in URLs as URL-safe Base64.
 const BLOB_ID = bcs.u256().transform({
     input: (id: string) => id,
     output: (id) => base64UrlSafeEncode(bcs.u256().serialize(id).toBytes()),
 });
 
+// Different than BLOB_ID, because we don't want this to be URL-safe;
+// otherwise, it will mess up with the checksum results.
+const U256 = bcs.u256().transform({
```
giac-mysten, line 21: Nit: something like this?
```suggestion
const DATA_HASH = bcs.u256().transform({
```

giac-mysten (approved):
Thanks for the changes @Tzal3x ! LGTM!

portal/common/lib/crypto.ts
Tzal3x: Yes it works, here are the docks for nodejs: https://nodejs.org/api/webcrypto.html#cryptosubtle 

The server portal preview deployment is also working, so that's a proof too.

portal/common/lib/bcs_data_parsing.ts
Tzal3x: Done ✔️

Issue #210, opened 2024-09-16
Smart contract & common portal lib: support any HTTP header.


Tzal3x:

Tzal3x:
📝 A note on why we need blob hashing along with blobId. You can think of it as a very structured hash of the blob (which also depends on the encoding of the slivers 👉 [source](https://github.com/MystenLabs/walrus/blob/main/docs/red-stuff.md#sliver-authentication-blob-metadata-and-the-blob-id)). 

Computing the blobId from the blob contents to check if the aggregator is trustworthy is computationally expensive to happen on the client side. 

On the other hand, just hashing the contents returned by the aggregator, instead of regenerating the blobId is faster.
So we need to add that `blob_hash` to the `site::Resource` to accomplish the above.

PR #221, opened 2024-09-24
feat: support any http header

Tzal3x:
### Summary

Related to https://github.com/MystenLabs/walrus-sites/issues/210

From now on each ws site::Resource can contain any type of http response header.
Http headers are specified in a `ws-responses.json` file which should be contained in the root directory of the folder that contains the files to be uploaded on walrus (usually `/dist`). 
The file can also be specified using the CLI. 

If the content-type and content-encoding headers are not specified explicitly, they are deduced using the file extension. 
Otherwise defaults are used. 

### TODO
- [x] Remove `content-type` and `content-encoding` fields from site::Resource. Http headers should no longer be hard-coded, since we will now support **any** header.
- [x] Site builder: Parse http headers from `ws-config.json` and add them to the `site::Resource::headers`.
- [x] On update of `ws-config.json`, make sure the Resource headers get updated. 
- [x] Rename `ws-config.json` to `ws-resources.json`
- [x] Use VecMap instead of Vector for the `site::Resource.headers`
- [x] Portal: extract the headers from the `site::Resource` and apply them correspondingly. #229


Tzal3x:
Facing issue executing the PTB: 
`Error: error while processing the Sui transaction: CommandArgumentError { arg_idx: 1, kind: InvalidUsageOfPureArg } in command 1`

site-builder/src/site/resource.rs
```diff
@@ -348,22 +396,42 @@ impl ResourceManager {
                 .expect("the path should not terminate in `..`"),
```
giac-mysten, line 396: Question (line 391): should we remove the content encoding as a parameter, and rely on the headers?

site-builder/src/site/resource.rs
```diff
@@ -384,8 +452,7 @@ impl ResourceManager {
         Ok(Some(Resource::new(
             full_path_to_resource_path(full_path, root)?,
             full_path.to_owned(),
-            content_type,
-            *content_encoding,
+            http_headers.unwrap(),
```
giac-mysten, line 455: Important: use `except("explanation why bad stuff should not happen")` instead of `unwrap`
```suggestion
            http_headers.except("we have created the headers above"),
```

Nit: removing TryFrom also removes the need for an expect here.

site-builder/src/site/resource.rs
```diff
@@ -348,22 +396,42 @@ impl ResourceManager {
...
+        let headers = match all_headers {
+            Some(map) => map.get(&resource_path),
+            _ => None,
+        };
+        // 3. Does the specific resource has headers defined for it? If not, use defaults.
+        let http_headers = match headers {
+            Some(h) => HttpHeaders::try_from(h.clone()),
+            _ => Ok(HttpHeaders::new()),
+        };
+
```
giac-mysten, line 417: Important: I think that, if there are no content type headers, we should infer it from the extension

giac-mysten (commented):
Thank you @Tzal3x ! first batch of comments from my side :)

site-builder/src/site/resource.rs
```diff
@@ -323,13 +368,16 @@ pub(crate) struct ResourceManager {
     pub walrus: Walrus,
     /// The resources in the site.
     pub resources: ResourceSet,
+    /// The ws-config.json contents.
+    pub ws_config: Option<WSConfig>,
```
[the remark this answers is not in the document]
Tzal3x: I defined this as optional because it is not required to specify a `ws_config` for every walrus site. If it's not an option, then ws_config should always be contained in the user's folder right?  

Having optional fields on the other hands, implies that the file is defined, but some of the fields are missing.

site-builder/src/site/resource.rs
```diff
@@ -348,22 +396,42 @@ impl ResourceManager {
...
-                    );
-                    ContentType::ApplicationOctetstream // arbitrary binary data RFC 2046
-                }
-            };
+        // Get the headers of the specific resource if they exist
+        // for this specific path.
+        // 1. Is ws-config.json specified?
+        let all_headers = match &self.ws_config {
+            Some(config) => &config.headers,
+            None => return Ok(None),
```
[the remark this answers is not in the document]
Tzal3x: Ah, I thought that the return would only propagate to `let all_headers = Ok(None)` but as it seems it applies for the whole function. 🤦‍♂️ 

I will update this asap!

site-builder/src/site/resource.rs
```diff
@@ -56,14 +56,61 @@ impl TryFrom<&SuiMoveStruct> for ResourceInfo {
...
+#[derive(serde::Serialize, Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
+pub struct HttpHeaders(pub Vec<HttpHeader>);
+
+impl HttpHeaders {
+    pub fn new() -> Self {
+        HttpHeaders(Vec::new())
+    }
+}
+
+impl TryFrom<HashMap<String, String>> for HttpHeaders {
```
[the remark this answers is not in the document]
giac-mysten: Nvm, it `From` is simpler.

site-builder/src/site/resource.rs
```diff
@@ -56,14 +52,59 @@ impl TryFrom<&SuiMoveStruct> for ResourceInfo {
...
+impl From<HashMap<String, String>> for HttpHeaders {
+    fn from(values: HashMap<String, String>) -> HttpHeaders {
+        let mut headers = Vec::new();
+        for value in values {
+            headers.push(HttpHeader {
+                name: value.0,
+                value: value.1,
+            });
+        }
+        HttpHeaders(headers)
```
giac-mysten, line 86: Nit: slightly more idiomatic:
```suggestion
    fn from(values: HashMap<String, String>) -> Self {
        Self(
            values
                .into_iter()
                .map(|(name, value)| HttpHeader { name, value })
                .collect(),
        )
    }
```

site-builder/src/site/resource.rs
Tzal3x: Yes! Content encoding remained there only temporarily in order to not break stuff. 
Now it has been removed from the parameters. ✔️

giac-mysten (approved):
Great, thanks so much @Tzal3x ! let's merge this and iterate in follow ups

PR #226, opened 2024-09-30
feat(site-builder): Add --version support

mario4tier:
**Goal**
Have the following to work same as the way the sui binary works:
```
site-builder --version 
site-builder -V
```

**Why?**
Help Suibase scripts to maintain latest binary versions for its devs.

**Implementation**
Trivial cut&paste of how sui binary already handles this (using Mysten Labs bin-version crate).

**Tests**
image

### DIFF
--- a/site-builder/src/main.rs
+++ b/site-builder/src/main.rs
@@ -21,7 +21,7 @@ use crate::{
     util::{get_existing_resource_ids, id_to_base36, load_wallet_context},
 };
 
-// Define the `GIT_REVISION` and `VERSION` consts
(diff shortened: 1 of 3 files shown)


site-builder/src/main.rs
```diff
@@ -21,8 +21,11 @@ use crate::{
     util::{get_existing_resource_ids, id_to_base36, load_wallet_context},
 };
 
+// Define the `GIT_REVISION` and `VERSION` consts
```
giac-mysten, line 24: Nit:
```suggestion
// Define the `GIT_REVISION` and `VERSION` consts.
```

PR #227, opened 2024-10-01
Mock System Calls to Reduce Benchmark Variance

chariskms:


portal/common/lib/page_fetching.bench.ts
```diff
@@ -1,20 +1,132 @@
...
+        vi.mock('./bcs_data_parsing', () => ({
+            ResourcePathStruct: {},
+            ResourceStruct: {},
+            DynamicFieldStruct: () => ({
+                parse: vi.fn().mockReturnValue({
+                    value: {
+                        blob_id: '0xresourceBlobId',
+                        path: '/index.html',
+                        content_type: 'text/html',
+                        content_encoding: 'utf8',
```
Tzal3x, line 62: Very soon this will need to be replaced by `headers` because we will accept any kind of http header using a `map<String,String>` so this mock will be deprecated by EOW. 

To avoid double work, I would suggest rebasing from develop as soon as #210 gets merged and then make this change there.

portal/common/lib/resource.bench.ts
```diff
@@ -1,21 +1,74 @@
...
+vi.mock('./bcs_data_parsing', () => ({
+    ResourcePathStruct: vi.fn(),
+    ResourceStruct: vi.fn(),
+    DynamicFieldStruct: vi.fn(() => ({
+        parse: vi.fn().mockReturnValue({
+            value: {
+                blob_id: '0xresourceBlobId',
+                path: '/index.html',
+                content_type: 'text/html',
+                content_encoding: 'utf8',
```
Tzal3x, line 28: same as comment above

portal/common/lib/resource.bench.ts
```diff
@@ -1,21 +1,74 @@
...
             const resourcePath = '/index.html';
-            const resp = await fetchResource(client, objectId, resourcePath, new Set());
+
+            // Use the mocked client and call fetchResource.
+            const resp = await fetchResource(mockClient, objectId, resourcePath, new Set());
+
+            // Validate that the fetched resource is versioned.
             expect(isVersionedResource(resp)).toBeTruthy();
         });
-    })
```
Tzal3x, line 20: `SITES_USED_FOR_BENCHING`: This will no longer be needed. We used it when we made real network requests to the actual sites. 

Now there should just be:
1. A test for an occasion like the landing page (with no redirects).
2. A test for a page **with** redirects (such as accessing a Flatlander).

Tzal3x (changes requested):
Nice mock-fu @chariskms! 🙇‍♂️ Requested some additional stuff.

Tzal3x (approved):
Thank you @chariskms! 🚀

PR #232, opened 2024-10-03
chore(site-builder): small improvements and fixes

giac-mysten:
A list of many small changes:

* Cleans up some unused structs, removes unused dependencies.
* Bumps the sui versions, prevents cargo deny from complaining.
* Uniforms comments.
* Fetches the correct configuration file from the root site directory, and does not upload it with the site.
* **Does not reupload unchanged resources** - there was a bug in the comparison of the headers.


Tzal3x (approved):
🔥

PR #233, opened 2024-10-05
feat(site-builder/contracts): add possiblity to specify the default routes for a site

giac-mysten:
* upgrades the contracts to move 2023
* improves the diffing between the on chain site and the local directory
* adds routes to the contracts and the site builder config; the routes are stored in a VecMap in the `routes` dynamic field.


Tzal3x (approved):
Great rust-fu! 🥋 🔥

PR #234, opened 2024-10-07
feat(contract): Add range field to Resource

Tzal3x:
This is used when multiple resources
share the same blob. To extract each
resource, the byte range field specifies
the slice of the blob to retrieve.

The aim of this change is to reach a stable 
smart contract version upon walrus testnet release. 

The corresponding site-builder and portal 
changes will be included in a separate PR.


move/walrus_site/Move.lock
```diff
@@ -22,7 +22,7 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.34.2"
+compiler-version = "1.33.1"
```
giac-mysten, line 25: Question: is this intentional?

move/walrus_site/sources/site.move
```diff
@@ -26,7 +28,12 @@ module walrus_site::site {
         blob_id: u256,
         // Contains the hash of the contents of the blob
         // to verify its integrity.
-        blob_hash: u256
+        blob_hash: u256,
+        // Defines the byte range of the resource contents
+        // in the case where multiple resources are stored
+        // in the same blob. This way, each resource will
+        // be parsed using its' byte range in the blob.
+        range: Option<vector<u256>>,
```
giac-mysten, line 36: Question: why use an `Option<vector<u256>>`? Can we use tuples? I don't remember if move allows them. 

Suggestion: If it does not, we could strongly type it instead as `Range { start: Option<u256>, end: Option<u256> }`. Then, add an `Option<Range>` to the `Resource`. This may be a bit more readable and easy to parse? And also we don't need the length checks.

move/walrus_site/Move.lock
Tzal3x: No, because I didn't rebase the PR 😇

Tzal3x: Just rebased, should be fixed now.

move/walrus_site/sources/site.move
Tzal3x: Yes as far as I know there is no tuple definition in move. Sure I can use a Range struct 🙆‍♂️ !

move/walrus_site/sources/site.move
```diff
@@ -53,13 +65,40 @@ module walrus_site::site {
...
+        range_start: Option<u256>,
+        range_end: Option<u256>
     ): Resource {
+        let start_is_defined = option::is_some(&range_start);
+        let end_is_defined = option::is_some(&range_end);
+        // If defined, upper range bound should be positive (zero excluded).
+        if (end_is_defined) {
+            let end = option::borrow(&range_end);
+            assert!( *end > 0, ERangeEndNotPositive);
+        };
```
giac-mysten, line 78: Remark: `u256` cannot be negative.

Question: is this to avoid ranges of length 0?

move/walrus_site/sources/site.move
```diff
@@ -53,13 +65,40 @@ module walrus_site::site {
...
+        // If both range bounds are defined, the upper bound should be greater than the lower.
+        if (start_is_defined && end_is_defined) {
+            let start = option::borrow(&range_start);
+            let end = option::borrow(&range_end);
+            assert!(*end > *start, ERangeStartGreaterThanRangeEnd);
```
giac-mysten, line 83: Question: just making sure that we know we are not allowing ranges of size 0

move/walrus_site/tests/site_tests.move
```diff
@@ -0,0 +1,79 @@
...
+    }
+    #[test]
+    #[expected_failure(abort_code = ERangeEndNotPositive)]
+    fun test_new_resource_range_upper_bound_cannot_be_zero() {
+        walrus_site::site::new_resource(
+            b"/index.html".to_string(),
+            12121212,
+            89898989,
+            option::none(),
+            option::some(0)
```
giac-mysten, line 65: Remark: this is a bit weird: `some(0), none` (full range)  is allowed, but `none, some(0)` (empty range)

move/walrus_site/sources/site.move
```diff
@@ -53,13 +65,40 @@ module walrus_site::site {
...
+        let range: Option<Range> = if (!start_is_defined && !end_is_defined) {
+            option::none()
+        } else {
+            option::some(
+                Range {
+                    start: range_start,
+                    end: range_end
+                }
+            )
+        };
```
giac-mysten, line 95: Suggestion: separate this into a `Range::new` function.

move/walrus_site/sources/site.move
```diff
@@ -53,13 +65,40 @@ module walrus_site::site {
     public fun new_resource(
         path: String,
         blob_id: u256,
-        blob_hash: u256
+        blob_hash: u256,
+        range_start: Option<u256>,
+        range_end: Option<u256>
```
giac-mysten, line 70: The other option would be to use a ptb and pass here `range: Option<Range>` directly, which is constructed in a `Range::new` function.

move/walrus_site/sources/site.move
```diff
@@ -26,7 +28,17 @@ module walrus_site::site {
...
-        blob_hash: u256
+        blob_hash: u256,
+        // Defines the byte range of the resource contents
+        // in the case where multiple resources are stored
+        // in the same blob. This way, each resource will
+        // be parsed using its' byte range in the blob.
+        range: Option<Range>,
+    }
+
+    public struct Range has store, drop {
```
giac-mysten, line 39: Question: does this need `store`?
```suggestion
    public struct Range has drop {
```

giac-mysten (commented):
Thanks for the changes! a few more suggestions

move/walrus_site/sources/site.move
Tzal3x: Yes it is done for checking zero lengths!

move/walrus_site/tests/site_tests.move
Tzal3x: `none, some(0)` is not allowed, as it would imply an empty range, correct. But what is the problem?

move/walrus_site/sources/site.move
Tzal3x: Yea the `store` ability is required for every type in order to be stored in objects with the key ability (Resource).

Tzal3x: https://move-book.com/storage/store-ability.html

move/walrus_site/tests/site_tests.move
Tzal3x: After our call: I will enable support for empty files, since there are cases where there are empty files in folders!

Issue #220, opened 2024-09-24
Support path prefix matching for client-side routing


Tzal3x:
In `ws-resources.json`, define routes pointing to other routes for [client-side routing](https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating):

```
{
  ...
  "routes": [
    {
      "route": "/*",
      "serve": "/index.html",
    },
    ...
  ]
  ...
}
```

The final goal is to specify routes like it's done in the [nextjs middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware).

For starters, we can avoid complex expressions like `/some/*/path` and only implement patterns that end with wildcards like this: `/some/path/*`.

**Implementation**

* Each Walrus site requires its own `ws-resources.json` in site resources.
* When a walrus site is fetched for the first time, the portal should try to also parse the `ws-resources.json` resource.

When present, the portal prioritizes routing requests based on this configuration file before falling back to default routing mechanisms.

PR #235, opened 2024-10-07
feat(portal): integrate resource routing to portal

Tzal3x:
Update the portal code so that it aligns with the changes made in #220

Changelog:

- [x]  Create a `routing.ts` file containing the logic of parsing the Routes dynamic field.
- [x]  Add a function that finds if a path matches a route specified in the Routes DF.


portal/common/lib/page_fetching.ts
```diff
@@ -28,7 +29,19 @@ export async function resolveAndFetchPage(parsedUrl: DomainDetails): Promise<Res
...
+        if (!routes) {
+            console.warn("No routes found for the object ID");
+            return fetchPage(client, resolveObjectResult, parsedUrl.path);
+        }
+        let matchingRoute: string | undefined;
+        matchingRoute = matchPathToRoute(parsedUrl.path, routes)
+        if (!matchingRoute) {
+            console.warn(`No matching route found for ${parsedUrl.path}`);
+        }
+        return fetchPage(client, resolveObjectResult, matchingRoute ?? parsedUrl.path);
```
giac-mysten, line 44: Remark: seems to me here that we first try the routes, and then if there are no routes we try to get the object at the direct path. This may be problematic (imagine if we have an `index.html` with a `/*` route, but still have some other static assets like `/image.png`. we would never be able to load the image). 

Also, this is contrary to the _longest prefix matching_ that we do on the routes (i.e., the longest possible prefix is the full path, so we should match that first). 

Suggestion: if the above is correct, I would suggest to change to first trying the full path, and if that fails match the route. Maybe also start fetching the routes in the backgroud while the full path is fetched.

portal/common/lib/page_fetching.ts
```diff
@@ -70,6 +83,8 @@ export async function fetchPage(
     objectId: string,
     path: string,
 ): Promise<Response> {
+    // TODO: On the portal side, fetch the route object, store it to a routes object.
```
giac-mysten, line 86: Question: is the todo still valid? if it is, link it to an issue.

portal/common/lib/routing.test.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const testCases = [
+    ["/path/to/somewhere/", "/somewhere.html"],
+    ["/somewhere/else", 'else.jpeg'],
+    ["/", "/default.html"],
+    ["/somewhere", "/default.html"],
+    ["/somewhere/else/star", "star-else.gif"],
+    ["/somewhere/else/", 'star-else.gif'],
+]
```
giac-mysten, line 35: Remark: note that in our syntax a valid resource path _must_ always start with `/`. So, e.g., `else.jpg -> /else.jpg`

portal/common/lib/routing.ts
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Matches the path to the appropriate route.
+ * Path patterns in the routes list are sorted by length in descending order.
+ * Then the first match is returned.
+ *
+ * @param path - The path to match.
+ * @param routes - The routes to match against.
+ */
+export function matchPathToRoute(path: string, routes: Routes): string | undefined {
+    // TODO: improve this using radix trees.
```
giac-mysten, line 93: Remark: radix trees only improve on a linear pass only if we do some caching.

portal/common/lib/page_fetching.ts
Tzal3x: No, that is already done ✅

Tzal3x: Removed.

portal/common/lib/routing.ts
Tzal3x: We do implement a cache for routes objects here: https://github.com/MystenLabs/walrus-sites/pull/236

In the furure, whenever the matchPathToRoute is called we could update the Routes object in the cache and get the benefits of radix trees.

portal/common/lib/page_fetching.ts
Tzal3x: Based on the [tests](https://github.com/MystenLabs/walrus-sites/blob/e86a2b6a5a76bf59ab1e37bad2c21635772569bf/portal/common/lib/routing.test.ts#L19-L35) this case is already covered and the longest prefix matching applies. 

i.e. given `/else.jpeg` the longest prefix will match first, i.e. both `/somewhere/else/*` and `/somewhere/else` will be checked first, and then since no matches are found in the specified routes, the last option is the fallback of `['/*', '/default.html']` which means _"catch everything else and redirect it to default"_. 

If `/else.jpeg` doesn't match the `/default.html`, it is a dev misconfiguration, isn't it?

giac-mysten: quick example to see if we are saying the same thing or not :)

let's say i have files:
- `/index.html`
- `/image.jpg`

and routes: `/* : /index.html`

i would expect this to work, and that all paths _except_ `/image.jpg`are routed to `/index.html`.

I.e., I would first check if there is a direct match, and if there isn't, check the routes.

Tzal3x: oh! I see, then yes this is not happening. I can make this work.

giac-mysten: but do you agree this should be the expected behavior? I'm not too sure what happens e.g., with vercel json

Tzal3x: It's a matter of taste. [Nextjs middleware matcher](https://nextjs.org/docs/app/building-your-application/routing/middleware) would match the pattern as I have implemented it initially.

The rule of  checking first for `/image.jpg` and if it's not found, then try to apply re-route to `/index.html` based on pattern matching is a rule that is implied, and I wouldn't say that it would be an expected behavior.

Meaning, if I have specified a pattern, to reroute to a specific path, then I would first expect for my rule to apply.

portal/common/lib/routing.ts
```diff
@@ -0,0 +1,101 @@
...
+    const sortedRoutes: Array<[string, string]> = Array.from(
+        routes.routes_list.entries()
+    ).sort((current, next) => next[0].length - current[0].length);
+    const matchedRoute = sortedRoutes.find(
+        ([pattern, _]) => new RegExp(`^${pattern.replace('*', '.*')}$`).test(path)
+    );
```
Tzal3x, line 99: Implement this without sort since we can iterate once and get the longest path match. O(n) instead. 👍

portal/common/lib/page_fetching.ts
Tzal3x: For future us: After our call we concluded that it would be best to check first for a direct path match, and if no resource was fetched, then we can use the Routes object to resolve it. This would be compatible with SPAs.

portal/common/lib/routing.ts
```diff
@@ -89,13 +89,11 @@ function parseRoutesData(bcsBytes: string): Routes {
...
-    ).sort((current, next) => next[0].length - current[0].length);
-    const matchedRoute = sortedRoutes.find(
-        ([pattern, _]) => new RegExp(`^${pattern.replace('*', '.*')}$`).test(path)
-    );
-    return matchedRoute? matchedRoute[1] : undefined
+    const res = Array
+        .from(routes.routes_list.entries())
+        .filter(([pattern, _]) => new RegExp(`^${pattern.replace('*', '.*')}$`).test(path))
+        .reduce((a, b) => a[0].length >= b[0].length ? a : b)
+    return res ? res[1] : undefined;
```
giac-mysten, line 98: Question: just to check, `res[1]` is to get the resource for the path that matched, right?

portal/common/lib/routing.ts
```diff
@@ -89,13 +89,11 @@ function parseRoutesData(bcsBytes: string): Routes {
...
-        routes.routes_list.entries()
-    ).sort((current, next) => next[0].length - current[0].length);
-    const matchedRoute = sortedRoutes.find(
-        ([pattern, _]) => new RegExp(`^${pattern.replace('*', '.*')}$`).test(path)
-    );
-    return matchedRoute? matchedRoute[1] : undefined
+    const res = Array
+        .from(routes.routes_list.entries())
+        .filter(([pattern, _]) => new RegExp(`^${pattern.replace('*', '.*')}$`).test(path))
+        .reduce((a, b) => a[0].length >= b[0].length ? a : b)
```
giac-mysten, line 97: Praise: beautiful!

giac-mysten (approved):
Awesome stuff, thanks @Tzal3x !

portal/common/lib/routing.ts
Tzal3x: Correct!

PR #236, opened 2024-10-08
feat(portal): add caching interface to sw portal routing

Tzal3x:
Adds caching to the routing logic. 

- [x] Implement the routing cache interface for the service worker portal
- [ ] Implement the routing cache interface for the server portal

### DIFF
--- a/examples/snake/ws-resources.json
+++ b/examples/snake/ws-resources.json
@@ -6,8 +6,6 @@
     }
   },
   "routes": {
-    "/path/*": "/index.html",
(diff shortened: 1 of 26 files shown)


Tzal3x:
Open questions: 
- Expiration date. 
- Update Routes when `ws-resources.json` changes. 
- Does a new db get created for every walrus site?

Tzal3x:
Closing as it has diverged significantly from `main`, plus it might not be worth it introducing more complexity for little perf gains.

PR #237, opened 2024-10-08
feat(service-worker): handle cache storage overflow

Tzal3x:
In case the cache storage of the service worker if full, a QuotaExceededError will be raised. 

To handle it, I delete the oldest 50 entries in the SW storage.

### DIFF
--- a/portal/worker/src/caching.ts
+++ b/portal/worker/src/caching.ts
@@ -32,7 +32,26 @@ export default async function resolveWithCache(
     console.log("Cache miss!", urlString);
     const resolvedPage = await resolveAndFetchPage(parsedUrl);
 
-    cache.put(urlString, resolvedPage.clone());
+    try {
+        await cache.put(urlString, resolvedPage.clone());
+    } catch (e) {
+        if (e.name === 'QuotaExceededError') {
+            console.warn("Cache quota exceeded, unable to store the response:", e);
+            const keys = await cache.keys();
+            // Delete at most N oldest entries.
+            for(let i = 0; i < 50; i++) {
+                if (i > keys.length) break;
+                const oldestKey = keys[i];
+                await cache.delete(oldestKey);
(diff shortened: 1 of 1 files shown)


portal/worker/src/caching.ts
```diff
@@ -32,7 +32,26 @@ export default async function resolveWithCache(
...
+            const keys = await cache.keys();
+            // Delete at most N oldest entries.
+            for(let i = 0; i < 50; i++) {
+                if (i > keys.length) break;
+                const oldestKey = keys[i];
+                await cache.delete(oldestKey);
+                console.log('Deleted oldest cache entry:', oldestKey);
+            };
+            // Retrying...
+            await cache.put(urlString, resolvedPage.clone());
```
giac-mysten, line 49: Question: what happens if you fail again?

portal/worker/src/caching.ts
```diff
@@ -32,7 +32,26 @@ export default async function resolveWithCache(
...
+    try {
+        await cache.put(urlString, resolvedPage.clone());
+    } catch (e) {
+        if (e.name === 'QuotaExceededError') {
+            console.warn("Cache quota exceeded, unable to store the response:", e);
+            const keys = await cache.keys();
+            // Delete at most N oldest entries.
+            for(let i = 0; i < 50; i++) {
+                if (i > keys.length) break;
+                const oldestKey = keys[i];
```
giac-mysten, line 44: Question: are keys sorted by index in the cache?

portal/worker/src/caching.ts
giac-mysten: One way to deal with this would be to separate it to a function and then call itself recursively, such that more and more elements are deleted if the `put`s fail

giac-mysten (approved):
Thanks for the changes, awesome!

PR #238, opened 2024-10-09
feat(site-builder): new unchanged resource operation

chariskms:
In this PR, we address the issue of potential missing blobs on Walrus due to devnet wipes. We ensure that resources in the Sui object that appear unchanged are re-uploaded to Walrus if necessary, preventing discrepancies between on-chain data and Walrus storage.

**Changes:**
Unchanged Resource Operation:

-  Added ResourceOp::Unchanged to handle resources that appear unchanged but may still need re-uploading.
-  Calculate unchanged resources inside diff function.
-  Modified publish_to_walrus to process both Created and Unchanged resources.

### DIFF
--- a/site-builder/src/site/builder.rs
+++ b/site-builder/src/site/builder.rs
@@ -104,6 +104,7 @@ impl SitePtb<Argument> {
             match call {
                 ResourceOp::Deleted(resource) => self.remove_resource_if_exists(resource)?,
                 ResourceOp::Created(resource) => self.add_resource(resource)?,
+                ResourceOp::Unchanged(_) => (),
             }
         }
         Ok(())
--- a/site-builder/src/site/manager.rs
+++ b/site-builder/src/site/manager.rs
@@ -105,7 +105,7 @@ impl SiteManager {
     async fn publish_to_walrus<'b>(&self, updates: &[ResourceOp<'b>]) -> Result<()> {
         let to_update = updates
             .iter()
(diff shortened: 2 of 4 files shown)


giac-mysten (commented):
Very nice! thanks @chariskms for the iterations :) qq: did you try this out on a site? do the uploads trigger?

chariskms:
> Very nice! thanks @chariskms for the iterations :) qq: did you try this out on a site? do the uploads trigger?
Yes triggers store, only for the unchanged resources. Do you think we should do something about the 'force' option, such as removing it? I noticed that it has been changed to a when_upload.is_always() check. Do you think the best option to simulate exactly the situation we want is when a Walrus object becomes deletable?

giac-mysten:
> @giac-mysten Yes triggers the store only for the unchanged resources and gets as response AlreadyCertified. 

great!

> Do you think we should do something about the 'force' option, such as removing it? I noticed it has been changed to a when_upload. 

i think we should keep it, but move it to have the same meaning as with the walrus client: The force option ensures that a new blob resource on sui is created again on sui. but this can be done as a follow up.

> Also, do you think the best option to simulate the exact situation we want is when a Walrus blob becomes deletable?

mmm deletable blobs will need some more logic to be handled. What do you mean by "simulate the exact situation we want" :)

giac-mysten (approved):

Issue #35, opened 2024-06-05
Check the status of the blob on Walrus before performing operations.


giac-mysten:
The site builder currently only checks if the blob ID on chain matches the blob ID on disk.

However, if devnet was wiped, the two can be the same but the blob may be missing.

This is currently mitigated by the `--force` flag.

Tzal3x:
> if devnet was wiped, the two can be the same but the blob may be missing.

If devnet was wiped, how would one fetch the on-chain blobid in the first place to check the disk blob ID matches it?

giac-mysten:
If I understand your question, the point here it that "devnet" is the Walrus devnet. The "on-chain blob ID" is on Sui _testnet_. So the `Site` and relative `Resource` objects on chain may still contain a blob ID, but the storage nodes do not have it stored.

mlegner:
So the current behavior is the following:
- Without `--force`, do not attempt to store files on Walrus if the corresponding Walrus Sites `Resource` already exists.
- With `--force`, store all files on Walrus with the `--force` option (provided to the `walrus` binary).

What we would want to support the case that Walrus is redeployed is one of the following:
- Check the Walrus blob status for all files and store those that are not certified.
- Attempt to store all files on Walrus but _without_ using the `--force` option.

PR #241, opened 2024-10-10
feat(site-builder): parallelize calls to the walrus cli

giac-mysten:
Makes calling the site builder on large sites (tens/hundreds of files) much faster, especially in the case of small site updates. 

* parallelizes calls to the walrus CLI that can be made safely without locking objects.
* reduces the number of calls to the fullnodes by keeping the number of shards.
* only performs update checks when the command is a `--force`.

Example of improvement: a site with 50 files of 1KB each (`before -> after`):
- publish operation: `157s -> 102s`
- update with no changes: `67s -> 5s`
- update with 1-file change `70s -> 15s`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6474,6 +6474,7 @@ dependencies = [
  "libc",
  "mio 1.0.2",
  "pin-project-lite",
+ "signal-hook-registry",
  "socket2",
  "tokio-macros",
  "windows-sys 0.52.0",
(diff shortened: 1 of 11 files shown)


giac-mysten:
@chariskms this PR touches also the `Unchanged` resource operation. I realized that trying to re-upload unchanged resources as a way to achieve #35 is fine for small sites, but increases latency massively for big ones. One way to get around this problem is to go through the `blob-info` endpoint (instead of `store`), because it can be parallelized. 

Therefore, a solution could look like:
* find the set of unchanged resources (already implemented)
* check the blob status by parallel calls to the `blob-status` endpoint
* for the calls that return `nonexistent`, upload the blob again

however, one issue with this is that each of the calls requires a query to the full node. When done in parallel, these will quickly reach the rate limit for the public full node and the site builder will fail. 

So i'm not too sure what the best way forward is here. Thoughts?

Tzal3x (approved):
Nice 🚀

PR #242, opened 2024-10-11
feat(portal): propagate requests to a fallback devnet portal

Tzal3x:
Makes possible requests that are made on testnet, to be forwarded and retried on a portal that supports devnet. 
This way we can support old deployments that still have not migrated to testnet.

The [fallback devnet portal](https://vercel.com/mysten-labs/walrus-sites-sp-devnet-fallback) is a server side portal running the code of commit 3e693ad09d2670472e9d921b1dfab17166b6a9f6 (deployed using the branch [temp/devnet](https://github.com/MystenLabs/walrus-sites/tree/temp/devnet)). 

This is the state of our old walrus sites version (before reaching v.1.0.0), i.e. supporting only the devnet deployments, anything prior to the current date (2024/10/11).

### DIFF
--- a/portal/server/app/route.ts
+++ b/portal/server/app/route.ts
@@ -5,6 +5,7 @@ import { getDomain, getSubdomainAndPath } from '@lib/domain_parsing'
 import { redirectToAggregatorUrlResponse, redirectToPortalURLResponse } from '@lib/redirects'
 import { getBlobIdLink, getObjectIdLink } from '@lib/links'
 import { resolveAndFetchPage } from '@lib/page_fetching'
+import { HttpStatusCodes } from '@lib/http/http_status_codes'
 
 export async function GET(req: Request) {
     const originalUrl = req.headers.get('x-original-url')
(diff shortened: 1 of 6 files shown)


portal/server/app/route.ts
```diff
@@ -30,8 +31,20 @@ export async function GET(req: Request) {
     const requestDomain = getDomain(url)
 
     if (requestDomain == portalDomain && parsedUrl && parsedUrl.subdomain) {
-        console.log('fetching from the service worker')
-        return resolveAndFetchPage(parsedUrl)
+        const forwardToFallback = async () => {
+            const subdomain = parsedUrl.subdomain;
+            const fallbackDomain = process.env.FALLBACK_DEVNET_PORTAL;
+            return fetch(`https://${subdomain}.${fallbackDomain}${parsedUrl.path}`)
```
giac-mysten, line 37: Question: move to a separate function? or any specific reason it's in a closure?

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -48,16 +49,39 @@ self.addEventListener("fetch", async (event) => {
...
-            // When not being in a secure context, the Cache API is not available.
-            console.warn("Cache API not available");
-            response = resolveAndFetchPage(parsedUrl);
-        } else {
-            response = resolveWithCache(parsedUrl, urlString);
-        }
-        event.respondWith(response);
+    if (requestDomain === portalDomain && parsedUrl && parsedUrl.subdomain) {
+        // Fetches the page resources.
+        const handleFetchRequest = async (): Promise<Response> => {
```
giac-mysten, line 54: Suggestion: separate this out a bit for readability?

giac-mysten (approved):
LGTM!

portal/server/app/route.ts
Tzal3x: I preferred the closure since I could use outer scoped values inside it without creating function arguments, plus defining it the same place that it's used, keeping it simpler. 
It's also a short implementation and it will only be used temporarily, so when the time comes to remove it I will just delete this specific line range.

portal/worker/src/walrus-sites-sw.ts
Tzal3x: Done!

PR #243, opened 2024-10-11
feat(site-builder)!: implement resource ranges in the site builder

giac-mysten:


move/walrus_site/Move.lock
```diff
@@ -6,23 +6,23 @@ manifest_digest = "6955A2973AF0DB196E469AD72FDB6D203278DF6207C4561B7B9FDC4BD1AC9
...
 source = { git = "https://github.com/MystenLabs/sui.git", rev = "framework/testnet", subdir = "crates/sui-framework/packages/sui-framework" }
 
 dependencies = [
-  { id = "MoveStdlib", name = "MoveStdlib" },
+  { name = "MoveStdlib" },
 ]
 
 [move.toolchain-version]
-compiler-version = "1.35.0"
+compiler-version = "1.34.2"
```
Tzal3x, line 25: Is this one intentional?

site-builder/src/types.rs
```diff
@@ -119,3 +119,10 @@ impl RouteOps {
         matches!(self, RouteOps::Unchanged)
     }
 }
+
+/// Range of bytes for a resource.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord)]
+pub struct Range {
+    pub start: Option<u64>,
+    pub end: Option<u64>,
```
Tzal3x, line 127: Since the contract defines start and end as `u256`, why are we using u64 here?

giac-mysten: changed the contract as well. Since these are byte ranges, `64 bits` is more than enough: `2^64 bytes ~ 18 exabytes` 😅 we will never have blobs this big to index, and this way we save a bit of gas fees.

move/walrus_site/Move.lock
giac-mysten: good question, not sure 🤔

giac-mysten: reverted the lock file, I think it's just my compiler...

PR #245, opened 2024-10-11
feat(portal): add resource byte ranges to the portal

giac-mysten:


giac-mysten:
@Tzal3x this is still wip. Everything seems to work, still todo is to convert the numbers in the resource to http range headers in the request.

portal/common/lib/types/index.ts
```diff
@@ -25,50 +25,68 @@ export type Resource = {
...
     blob_hash: string;
+    range: Range | null;
 };
 
+export type Range = {
+    start: number | null;
+    end: number | null;
+};
+
+function rangeToHttpHeader(range: Range): string {
```
Tzal3x, line 36: Nit: I would add some sanity checks here regarding the values of start and end. 
e.g. if start and end defined, start>=0 end >=start

portal/common/lib/types/index.ts
```diff
@@ -25,50 +25,68 @@ export type Resource = {
...
         obj &&
-        typeof obj.path === 'string' &&
-        typeof obj.headers === 'object' &&
-        typeof obj.blob_id === 'string' &&
-        typeof obj.blob_hash === 'string'
+        typeof obj.path === "string" &&
+        typeof obj.headers === "object" &&
+        typeof obj.blob_id === "string" &&
+        typeof obj.blob_hash === "string" &&
+        typeof obj.range === "object"
```
Tzal3x, line 63: Note: we should fix the editor config to include one of either single or double quotes. 
This PR looks big because there are multiple cases like this 😅

portal/common/lib/routing.test.ts
```diff
@@ -2,41 +2,51 @@
...
 
 import { getRoutes, matchPathToRoute } from "./routing";
-import { test, expect } from 'vitest';
+import { test, expect } from "vitest";
 import { SuiClient, getFullnodeUrl } from "@mysten/sui/client";
 import { NETWORK } from "./constants";
 
-const snakeSiteObjectId = '0x3e01b1b8bf0e54f7843596345faff146f1047e304410ed2eb85d5f67ad404206';
-test.skip('getRoutes', async () => {
+const snakeSiteObjectId = "0x3e01b1b8bf0e54f7843596345faff146f1047e304410ed2eb85d5f67ad404206";
```
Tzal3x, line 9: Nit: Since the contract was updated, the snake site object Id should also be updated. 

Even though we skip this test we still use it occasionally for local development.

portal/common/lib/routing.test.ts
```diff
@@ -2,41 +2,51 @@
...
+});
+
+// Test in the case there are no routes.
+const emptyRoutes = { routes_list: new Map<string, string>() };
+
+testCases.forEach(([requestPath, _]) => {
+    test(`matchPathToRoute: empty routes for "${requestPath}"`, () => {
+        const match = matchPathToRoute(requestPath, emptyRoutes);
+        expect(match).toEqual(undefined);
+    });
```
Tzal3x, line 51: Nice 🥇

portal/common/lib/routing.ts
```diff
@@ -89,11 +87,16 @@ function parseRoutesData(bcsBytes: string): Routes {
...
+    }
+
     // TODO: improve this using radix trees.
-    const res = Array
-        .from(routes.routes_list.entries())
-        .filter(([pattern, _]) => new RegExp(`^${pattern.replace('*', '.*')}$`).test(path))
-        .reduce((a, b) => a[0].length >= b[0].length ? a : b)
+    const res = Array.from(routes.routes_list.entries())
+        .filter(([pattern, _]) => new RegExp(`^${pattern.replace("*", ".*")}$`).test(path))
+        .reduce((a, b) => (a[0].length >= b[0].length ? a : b));
```
Tzal3x, line 99: Nit: Was this done automatically by the IDE? I would say the first indentation was more readable.

Tzal3x (approved):
One small commit for develop, one giant leap for walrus sites! 👨‍🚀 

Left some optional nits, otherwise feel free to merge! 😁 

Thanks @giac-mysten !

portal/common/lib/types/index.ts
giac-mysten: i think the issue is that `pre-commit` was not run at times, and therefore the previous code was not consistently formatted with that. Agree that we should enforce this programmatically.

portal/common/lib/routing.ts
giac-mysten: yep, I think this is still the `pre-commit` hook.

portal/common/lib/types/index.ts
giac-mysten: uhm, these are values that are taken from chain. The checks are enforced there. Do we need them here too? 🤔

Tzal3x: While that is true, I would add the checks because common will be an sdk, and in the future one could use this function in a different context that bypasses those rules.

giac-mysten: updated in b72764b0629b4a21dd300734a0afc916cccaf5ef

PR #246, opened 2024-10-14
feat(common-lib): derive dynamic field instead of call to fullnode 

Tzal3x:
- Upgrade @mysten/sui to 1.12.0
This version includes the `deriveDynamicFieldID`
function needed to calculate the DF object ID
without making a request to a fullnode.
- Replace `client.getDynamicFieldObject` with `deriveDynamicFieldID`.
- Skip `page_fetching` benchmarks. The results we get from it are misleading and slow down development whenever there are changes happening in its' dependencies. In the near future we should revisit how to properly bench this. https://github.com/MystenLabs/walrus-sites/issues/250

TODO (following PR)
- ⚡ Performance improvement: Use `multiGetObjects` to fetch both the display and the dynamic field object in the same request.

### DIFF
--- a/portal/common/lib/resource.test.ts
+++ b/portal/common/lib/resource.test.ts
@@ -13,7 +13,6 @@ import { RESOURCE_PATH_MOVE_TYPE } from './constants';
 
 // Mock SuiClient methods
 const getObject = vi.fn();
-
 const mockClient = {
     getObject,
 } as unknown as SuiClient;
--- a/portal/common/lib/resource.test.ts
+++ b/portal/common/lib/resource.test.ts
@@ -12,11 +12,9 @@ import { DynamicFieldStruct } from './bcs_data_parsing';
 import { RESOURCE_PATH_MOVE_TYPE } from './constants';
 
 // Mock SuiClient methods
-const getDynamicFieldObject = vi.fn();
 const getObject = vi.fn();
 
 const mockClient = {
-    getDynamicFieldObject,
     getObject,
 } as unknown as SuiClient;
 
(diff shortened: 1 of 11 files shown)


portal/common/lib/resource.test.ts
```diff
@@ -190,53 +182,11 @@ describe('fetchResource', () => {
...
+        getObject.mockResolvedValueOnce(undefined);
 
-        const result = await fetchResource(mockClient, '0xParentId', '/path', seenResources);
+        const result = await fetchResource(mockClient, '0x1', '/path', seenResources);
 
         // Check that the function returns NOT_FOUND
         expect(result).toBe(HttpStatusCodes.NOT_FOUND);
     });
-
-    test('should correctly handle a chain of redirects', async () => {
```
giac-mysten, line 201: question: has this been removed?

portal/common/lib/resource.ts
```diff
@@ -41,38 +47,25 @@ export async function fetchResource(
...
     }
 
-    // Initiate a pre-fetch for the checkRedirect operation without resolving it.
-    // We don't need the result yet, but it's useful if we will need
-    // it later, so we don't have to loose time.
-    const checkRedirectPromise = checkRedirect(client, objectId);
+    let redirectId  = await checkRedirect(client, objectId);
+    if (redirectId) {
+        fetchResource(client, redirectId, path, seenResources, depth + 1)
+    }
```
giac-mysten, line 53: Question/important: i think we still want to do the background prefetch of the redirect, so that we first check if there is a resource, and only if there isn't we redirect. Why the change?

portal/common/lib/resource.ts
```diff
@@ -41,38 +47,25 @@ export async function fetchResource(
...
     const pageData = await client.getObject({
-        id: dynamicFields.data.objectId,
+        id: dynamicFieldId,
         options: { showBcs: true },
     });
 
     // If no page data found.
-    if (!pageData.data) {
+    if (!pageData || !( pageData.data )) {
         console.log("No page data found");
```
giac-mysten, line 69: Suggestion: the `await` for the redirect could be here?

giac-mysten (approved):
Awesome, thanks @Tzal3x . Generally looks good, only one more important comment.

portal/common/lib/resource.test.ts
Tzal3x: Yes, I found out it was redundant.

PR #253, opened 2024-10-16
feat: move main to the testnet deployment!

giac-mysten:


Tzal3x (approved):
Fair winds and following seas!

PR #255, opened 2024-10-16
fix(portal): bug in redirects

giac-mysten:


portal/common/lib/resource.ts
```diff
@@ -62,10 +60,10 @@ export async function fetchResource(
     });
 
     // If no page data found.
-    if (!pageData || !( pageData.data )) {
+    if (!pageData || !pageData.data) {
         const redirectId = await redirectPromise;
         if (redirectId) {
-            fetchResource(client, redirectId, path, seenResources, depth + 1)
+            return fetchResource(client, redirectId, path, seenResources, depth + 1);
```
giac-mysten, line 66: this line was the problem, the rest is formatting...

Tzal3x (approved):
Nice find @giac-mysten!

PR #256, opened 2024-10-16
fix: add check that suins name has not changed to cache

giac-mysten:
The cache was not checking that the suins name for a site had not been changed. 
Therefore, if it was changed, the cache would not refresh.

The changes are a bit ugly and painful. I think it's soon going to be time to sit down and and improve the architecture of the worker and library.


giac-mysten:
@Tzal3x merging now w/o review to speed up testing. Please feel free to open another PR from this commit to add comments.

PR #259, opened 2024-10-17
fix: ensure the fallback returns uncompressed responses 

giac-mysten:
Additionally, show when the fallback is used also in prod.
Useful for checking if a site has been updated or not.


Tzal3x (approved):
@giac-mysten to the rescue! 🧯

PR #266, opened 2024-10-21
feat(site-builder): check in walrus config for sites-config.yaml

Tzal3x:


site-builder/src/main.rs
```diff
@@ -212,8 +212,22 @@ async fn run() -> Result<()> {
 
     let args = Args::parse();
     let mut config: Config = std::fs::read_to_string(&args.config)
+        .or_else(|err| {
+            if err.kind() == std::io::ErrorKind::NotFound {
+                let config;
+                if let Some(home_dir) = home::home_dir() {
+                    config =
+                        std::fs::read_to_string(home_dir.join(".walrus").join("sites-config.yaml"));
```
chariskms, line 220: I think, the CLI checks for the Walrus configuration file in the following order:

1. If a path is specified via the CLI option, it will attempt to read the config file from the given path. An error is returned if the path is invalid.
2. If no path is provided explicitly, the CLI will search for `client_config.yaml` or `client_config.yml` in the following locations:
   - The current working directory (`./`).
   - If the environment variable `XDG_CONFIG_HOME` is set, it will search in `$XDG_CONFIG_HOME/walrus/`.
   - If not, it will check in `~/.config/walrus/`.
   - Finally, it will check `~/.walrus/`.

I think that, as with client_config.yaml, it might be better to follow a similar approach here.

Tzal3x: Thank you @chariskms ! I will change this asap 😄

Tzal3x: Done! Uploaded and tested ✅ 
Everything works as expected.

Tzal3x: For future reference of @chariskms comment: https://github.com/MystenLabs/walrus/blob/main/crates/walrus-service/src/client/cli/args.rs#L23

site-builder/src/main.rs
```diff
@@ -212,8 +212,50 @@ async fn run() -> Result<()> {
 
     let args = Args::parse();
     let mut config: Config = std::fs::read_to_string(&args.config)
+        .or_else(|err| {
```
giac-mysten, line 215: Hint: this code can be made much smaller. I would borrow from the walrus cli code directly as in here. https://github.com/MystenLabs/walrus/blob/c567eef5179776756ea7b0d4c0c930830695c41a/crates/walrus-service/src/common/utils.rs#L513-L524

Tzal3x: Done ✅

site-builder/src/main.rs
```diff
@@ -206,12 +206,56 @@ mod default {
...
+            PathBuf::from(xdg_config_home)
+                .join("walrus")
+                .join("sites-config.yaml"),
+        );
+    }
+    // Define the home directory. If it doesn't exist, use `/`.
+    let home_dir = match home::home_dir() {
+        Some(dir) => dir,
+        None => PathBuf::from("/"),
+    };
```
giac-mysten, line 226: Question: why use root is home does not exist?

giac-mysten: i'd probably skip it if it is not there 🤔

giac-mysten (approved):
thanks @Tzal3x !

site-builder/src/main.rs
Tzal3x: If the home directory does not exist, I've seen systems that `$HOME` would then point to `/`. 

Hm, how would you skip it? 💭 
What would be the base of the other paths if no home dir is specified?

mario4tier:
IMHO, a command should fail with a "$HOME need to be defined" in very rare case of $HOME missing. That way the user knows that a fix is needed with **their** setup (cron/login/rlogin!?).

Defaulting to "/" may result in a more obscure error (e.g. a generic "access error")... and then you might have to provide support to explain "why?" Walrus is not able to load the files even if at one of the documented location.

site-builder/src/main.rs
giac-mysten: >Hm, how would you skip it? 💭

simply like this: https://github.com/MystenLabs/walrus/blob/c567eef5179776756ea7b0d4c0c930830695c41a/crates/walrus-service/src/common/utils.rs#L534-L536

if there is a home, look for the `~/.config...` path, otherwise don't add it to the list of path to search from.

giac-mysten:
> IMHO, a command should fail with a "$HOME need to be defined" in very rare case of $HOME missing. That way the user knows that a fix is needed with **their** setup (cron/login/rlogin!?).

i wouldn't fail here, only b/c there may just be a valid config in one of the other default paths. If none can be found, we can error out and print that no config was found across any of the default paths.

mario4tier:
The key is even if the file exists at one of the default path, it would fail to be seen by the command (because of a user **setup** error). That might result in an obscure support request to the Walrus team...

May be if the command reach the point of not finding a config anywhere... just additionally **inform** the user if $HOME is not defined?

(I don't feel strong about it, just a suggestion).

Tzal3x:
I also tend to agree that it would be a better approach throwing an error when `$HOME` is undefined. 

I just update the code, @giac-mysten if you agree, feel free to resolve the issue and then we can merge.

PR #267, opened 2024-10-21
feat(site-builder): remove epochs default value

Tzal3x:


site-builder/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "site-builder"
-version = "1.0.1"
+version = "1.0.2"
```
giac-mysten, line 3: Question: do we have any versioning strategy? 🤔

giac-mysten (approved):
🔥

site-builder/Cargo.toml
Tzal3x: I've been thinking of using something along those lines `MAJOR.MINOR.PATCH`.

PR #269, opened 2024-10-21
chore: delete benchmarks for resource and page_fetching

Tzal3x:


giac-mysten (commented):
Question: why delete the benches? I understand we do not want codespeed. But could these be useful outside that?

giac-mysten (approved):

Tzal3x:
> Question: why delete the benches? I understand we do not want codespeed. But could these be useful outside that?

The reason that the benchmarks are a pain to maintain because of the mocks, and they offer little to no value given that we already know our bottlenecks for page_fetching and resource, so they have become more like a decorative thing. 

Benchmarks on other files such as file decompression and pure functions are still nice to have, and they seem to be compatible with codspeed so I didn't remove them.

PR #272, opened 2024-10-22
feat: support wildcards in http headers

Tzal3x:
Checks the paths defined in `ws-resources.yaml` and if the resource path matches the pattern defined in the yaml, it adds the corresponding http headers to the `Resource::headers` vecmap.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -122,7 +122,7 @@ jobs:
         with:
           save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
       - name: Run tests
-        run: cargo test
+        run: cargo test -- --include-ignored
   check-all:
     name: Check if all code checks succeeded
     if: always()
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -122,7 +122,7 @@ jobs:
         with:
           save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
       - name: Run tests
-        run: cargo test -- --include-ignored
+        run: cargo test
   check-all:
     name: Check if all code checks succeeded
     if: always()
--- a/examples/snake/ws-resources.json
+++ b/examples/snake/ws-resources.json
@@ -3,6 +3,10 @@
     "/index.html": {
       "Content-Type": "text/html; charset=utf-8",
       "Cache-Control": "max-age=3500"
+    },
+    "/*.svg": {
+        "Cache-Control": "public, max-age=86400",
+        "ETag": "\"abc123\""
     }
   },
   "routes": {
--- a/site-builder/src/site/resource.rs
+++ b/site-builder/src/site/resource.rs
@@ -548,9 +548,24 @@ mod tests {
     }
 
     #[ignore = "The test depends on the file system containing a walrus binary.
-        Until we find a way to mock the walrus binary, this test will be ignored."]
+        // Until we find a way to mock the walrus binary, this test will be ignored."]
     #[tokio::test]
     async fn test_derive_http_headers() {
+        let resource_manager = setup_resource_manager_mock().await;
+
+        let resource_path = "/foo/bar/baz/image.svg";
+        let result = resource_manager.derive_http_headers(resource_path);
+
+        println!("Result: {:?}", result.keys());
+        assert_eq!(result.len(), 1);
+        assert_eq!(result.get("etag"), Some(&"\"abc123\"".to_string()));
+    }
+
+    /// Sets up a mock resource manager with the given headers configuration and resource path.
+    ///
+    /// Helper function for testing the `derive_http_headers` method.
+    async fn setup_resource_manager_mock() -> ResourceManager {
+        let walrus_mock = Walrus::new("walrus".to_string(), 1234, None, None, None);
         // Define the headers configuration for mocking the resource manager.
         let mut headers: BTreeMap<String, HttpHeaders> = BTreeMap::new();
         headers.insert(
@@ -567,23 +582,6 @@ mod tests {
                 "\"abc123\"".to_string(),
             )])),
         );
-        let resource_manager = setup_resource_manager_mock(headers).await;
-
-        let resource_path = "/foo/bar/baz/image.svg";
-        let result = resource_manager.derive_http_headers(resource_path);
-
-        println!("Result: {:?}", result.keys());
-        assert_eq!(result.len(), 1);
-        assert_eq!(result.get("etag"), Some(&"\"abc123\"".to_string()));
-    }
-
-    /// Sets up a mock resource manager with the given headers configuration and resource path.
-    ///
-    /// Helper function for testing the `derive_http_headers` method.
-    async fn setup_resource_manager_mock(
-        headers: BTreeMap<String, HttpHeaders>,
-    ) -> ResourceManager {
-        let walrus_mock = Walrus::new("walrus".to_string(), 1234, None, None, None);
         let ws_resources = Some(WSResources {
             headers: Some(headers),
             routes: None,
--- a/site-builder/src/site/resource.rs
+++ b/site-builder/src/site/resource.rs
@@ -567,10 +567,9 @@ mod tests {
                 "\"abc123\"".to_string(),
             )])),
         );
+        let resource_manager = setup_resource_manager_mock(headers).await;
 
         let resource_path = "/foo/bar/baz/image.svg";
-        let resource_manager = setup_resource_manager_mock(headers, resource_path).await;
-
         let result = resource_manager.derive_http_headers(resource_path);
 
         println!("Result: {:?}", result.keys());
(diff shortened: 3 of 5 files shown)


Tzal3x:
Main functionality is complete, now I should refactor the `http_headers` parsing into a separate function, and maybe add a few tests to it.

examples/snake/ws-resources.json
```diff
@@ -3,6 +3,10 @@
     "/index.html": {
       "Content-Type": "text/html; charset=utf-8",
       "Cache-Control": "max-age=3500"
+    },
+    "/*.svg": {
```
giac-mysten, line 7: Question: this is actually more generica than what we have for the routes. Do we want more expressive regexes such as this one? The routes only allow a wildcard at the end

site-builder/src/site/resource.rs
```diff
@@ -399,6 +384,43 @@ impl ResourceManager {
         )))
     }
 
+    /// Based on the ws-resources.yaml definition, this function derives the HTTP headers
+    /// based on wildcard matching, the headers that should be added to the HTTP response
+    /// of the resource defined by the resource_path.
```
giac-mysten, line 389: Suggestion: let's use the rustdoc convention for documentation here as well
```suggestion
    ///  Derives the HTTP headers for a resource based on the ws-resources.yaml.
    ///
    ///  Matches the path of the resource to the wildcard paths in the configuration to ....     
```

site-builder/src/site/resource.rs
```diff
@@ -399,6 +384,43 @@ impl ResourceManager {
...
+            })
+            .cloned()
+            // Cast the keys to lowercase because http headers
+            //  are case-insensitive: RFC7230 sec. 2.7.3
+            .map(|headers| {
+                headers
+                    .0
+                    .into_iter()
+                    .map(|(k, v)| (k.to_lowercase(), v))
+                    .collect()
```
giac-mysten, line 419: Suggestion: this could be done once when loading the configuration

site-builder/src/site/resource.rs
```diff
@@ -399,6 +384,43 @@ impl ResourceManager {
...
+            .and_then(|headers| {
+                headers
+                    .iter()
+                    .filter(|(path, _)| {
+                        // Replace the wildcard with a regex wildcard.
+                        let path_regex = path.replace('*', ".*");
+                        // Check if the resource_path matches the path_regex.
+                        match Regex::new(&path_regex) {
+                            Ok(re) => re.is_match(resource_path),
+                            Err(_) => false,
```
giac-mysten, line 403:
```suggestion
                        Regex::new(&path_regex)
                            .map(|re| re.is_match(resource_path))
                            .unwrap_or(false)
```

site-builder/src/site/resource.rs
```diff
@@ -399,6 +384,43 @@ impl ResourceManager {
...
+                        }
+                    })
+                    .max_by_key(|(path, _)| path.len())
+                    .map(|(_, header_map)| header_map)
+                    .into_iter()
+                    .reduce(|_, header_map| header_map)
```
giac-mysten, line 409: Question: why do you need the reduce here?

examples/snake/ws-resources.json
Tzal3x: In the case of headers there will be cases that the headers will be defined based on the file extension. 

Regarding routing, we use the same trick of replacing the wildcard `*` with the regex `.*` so this way of more expressive paths _is_ already supported there too.

site-builder/src/site/resource.rs
Tzal3x: After revisiting this it seems that is not needed! 😅 
Updated it, thanks for pointing it out

examples/snake/ws-resources.json
giac-mysten: >Regarding routing, we use the same trick of replacing the wildcard * with the regex .* so this way of more expressive paths is already supported there too.

of course

> So I would say this expressiveness would be needed.

makes sense.  However, we then have a problem with the "longest" matching. We cannot use the length of the string anymore.

for example, let's say we have:

- `/*.gnuplot`
- `/a/b/c/*`

both match `/a/b/c/plot.gnuplot`, and the former is a _shorter_ path, but actually it's a long string, so it matches first, leading to a bug in the behavior

Tzal3x: I updated the code so that there are unit tests to catch cases like the one that you discovered. 

To fix the longest path matching, I thought of including the matching pattern with the most slashes `'/'`:

```rust
matched_pattern.max_by_key(|(path, _)| path.split('/').count()) // Longest path match
```

The issue should be resolved now.

Tzal3x: It just came to mind that I didn't include the case of windows platforms that uses front slash instead of backslash. Need to account for this too!

giac-mysten: > It just came to mind that I didn't include the case of windows platforms that uses front slash instead of backslash. Need to account for this too!

uhm have you ever tried the site builder on windows? even without this modification, I'm not sure it would work.

site-builder/src/site/resource.rs
```diff
@@ -458,3 +487,118 @@ pub(crate) fn full_path_to_resource_path(full_path: &Path, root: &Path) -> Resul
...
+        path: &'static str,
+        expected: bool,
+    }
+
+    #[test]
+    fn test_is_pattern_match() {
+        let tests = vec![
+            PatternMatchTestCase {
+                pattern: "*.txt",
+                path: "file.txt",
```
giac-mysten, line 508: Question: shouldn't paths and patterns be all in the form `/some/pattern/*` and `/some/path` with a leading `/`?

site-builder/src/site/resource.rs
```diff
@@ -399,6 +399,35 @@ impl ResourceManager {
...
+                    .max_by_key(|(path, _)| path.split('/').count()) // Longest path match
+                    .map(|(_, header_map)| header_map.0.clone())
+            })
+            .unwrap_or_default()
+    }
+
+    /// Matches a pattern to a resource path.
+    ///
+    /// The pattern can contain a wildcard `*` which matches any sequence of characters.
+    /// e.g. `foo/*` will match `foo/bar` and `foo/bar/baz`.
```
giac-mysten, line 423: Nit: a resource path has a leading `/`.
```suggestion
    /// e.g. `/foo/*` will match `/foo/bar` and `/foo/bar/baz`.
```

.github/workflows/code.yml
```diff
@@ -122,7 +122,7 @@ jobs:
         with:
           save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
       - name: Run tests
-        run: cargo test -- --include-ignored
+        run: cargo test
```
giac-mysten, line 125: Suggestion: revert this and unignore the test (see later)

site-builder/src/site/resource.rs
```diff
@@ -458,3 +487,118 @@ pub(crate) fn full_path_to_resource_path(full_path: &Path, root: &Path) -> Resul
...
+            routes: None,
+        });
+        ResourceManager::new(
+            walrus_mock.clone(),
+            ws_resources,
+            Some(PathBuf::from(&"mock/path/to/ws-resources")),
+        )
+        .await
+        .unwrap()
+    }
```
giac-mysten, line 603: Hint: the function `derive_http_headers` only requires the `ws_resources` and the `resource_path`, and therefore should not require this complicated mock to test.  

Suggestion: make `derive_http_headers` a static function, that takes the two inputs above. Then, add a method to `ResourceManager` that calls that function. This way you can test the static function without the need to mock the manager.

site-builder/src/site/resource.rs
```diff
@@ -458,3 +487,118 @@ pub(crate) fn full_path_to_resource_path(full_path: &Path, root: &Path) -> Resul
...
+            "/foo/bar/baz/*.svg".to_string(),
+            HttpHeaders(BTreeMap::from([(
+                "etag".to_string(),
+                "\"abc123\"".to_string(),
+            )])),
+        );
+        let ws_resources = Some(WSResources {
+            headers: Some(headers),
+            routes: None,
+        });
```
giac-mysten, line 595: Suggestion: may be cleaner to provide a json an deserialize from it (see tests in `config.rs`)

giac-mysten (commented):
Thanks @Tzal3x ! a few comments from my side, but generally looks good.

site-builder/src/site/resource.rs
Tzal3x: Yes makes sense. Updated the tests. ✅

site-builder/src/site/resource.rs
Tzal3x: Nice idea, I avoided it initially because I didn't want to change the public interface of the ResourceManager but now that I think more about it, it makes sense. 💯

site-builder/src/site/resource.rs
Tzal3x: https://github.com/MystenLabs/walrus-sites/pull/272/commits/88be590bb0868b6785254444ec50cddc1ef1f701

site-builder/src/site/resource.rs
Tzal3x: https://github.com/MystenLabs/walrus-sites/pull/272#:~:text=action%20skip%20ignored-,6510a9d,-Refactor%3A%20mock%20headers

examples/snake/ws-resources.json
giac-mysten: >You mean that the site-builder on windows doesn't work in general?

yes, i don't think it does

giac-mysten (approved):
nice, thanks for the changes!

Issue #211, opened 2024-09-16
Add full-node fallback list


Tzal3x:
We could add load balancing routing on the portal side, i.e. monitor the available full nodes from it and pick the first available one in order of preference. 

Suggestion: add the possibility for users to pick their own trusted full node. 

Recommended full nodes:

```
Triton One
DSRV (All that node)
ANKR
Bware
Blockvision
```

mario4tier:
Sharing here that I have already develop a sui nodes proxy for network robustness (as part of a Sui foundation grant).

It is running by default for devs using Suibase. Your app just use a local daemon port and it "just work" (there is one port for testnet, another for mainnet etc...).

I understand you might prefer to have your own in-house solution, just a FYI.

What does it do?
================
- Configurable Full-nodes list (hot-reloaded without traffic disruption). 
- Periodic monitoring of configured nodes (e.g. average response time to a health check).
- Real-time "best nodes" subset for load-balancing (see "green" in picture).
- Automatic failover/retry, which is also "sui-aware" to prevent equivocation.
- Open-source (Rust, tokio optimized).

![image](https://github.com/user-attachments/assets/52a4f6da-5179-40bf-ada2-3478c8f1c771)

Tzal3x:
Thank you @mario4tier, we will consider your suggestion! 👍

PR #277, opened 2024-10-29
feat(portal)!: Add full-node fallback list by instantiating a sui client for each rpc node

chariskms:
This PR for Issue #211 adds a full-node fallback list to improve node reliability, enabling portal support for fallback nodes. It includes a fallback mechanism that creates a SuiClient instance for each RPC URL, ensuring resilience during network issues or rate limits.

**Changes:**
- RPC Selector: Introduced a singleton RPCSelector class that manages multiple SuiClient instances for efficient request handling across nodes.
- Fallback URLs: Added lists for testnetRPCUrls and prodRPCUrls to support testnet and mainnet failovers.
- API Refactor: Routed all SuiClient calls through RPCSelector to ensure reliable node selection during network issues or rate limiting.

Current RPC List:
  1. fullnode.testnet.sui.io
  2. sui-testnet.public.blastapi.io
  3. sui-testnet-endpoint.blockvision.org

  Not added due to a CORS error response in localhost tests:
   1. rpc.ankr.com/sui_testnet
   2. sui.blockpi.network/v1/rpc/public

### DIFF
--- a/portal/common/lib/constants.ts
+++ b/portal/common/lib/constants.ts
@@ -20,3 +20,5 @@ export const TESTNET_RPC_LIST = [
     'https://sui-testnet.public.blastapi.io',
     'https://sui-testnet-endpoint.blockvision.org'
 ];
+
+export const RPC_REQUEST_TIMEOUT_MS = 7000;
--- a/portal/common/lib/constants.ts
+++ b/portal/common/lib/constants.ts
@@ -14,3 +14,15 @@ export const SITE_NAMES: { [key: string]: string } = {
 export const FALLBACK_PORTAL = "blob.store";
 // The string representing the ResourcePath struct in the walrus_site package.
 export const RESOURCE_PATH_MOVE_TYPE = SITE_PACKAGE + "::site::ResourcePath";
+
+export const testnetRPCUrls = [
+    'https://fullnode.testnet.sui.io:443',
+    'https://sui-testnet.public.blastapi.io/',
+    'https://sui-testnet-endpoint.blockvision.org/',
+    'https://rpc.ankr.com/sui_testnet',
+];
+
+export const prodRPCUrls = [
+    'https://sui-mainnet-endpoint.blockvision.org/',
+    'https://sui-mainnet.public.blastapi.io/'
+];
--- a/portal/common/lib/constants.ts
+++ b/portal/common/lib/constants.ts
@@ -18,7 +18,5 @@ export const RESOURCE_PATH_MOVE_TYPE = SITE_PACKAGE + "::site::ResourcePath";
 export const TESTNET_RPC_LIST = [
     'https://fullnode.testnet.sui.io',
     'https://sui-testnet.public.blastapi.io',
-    'https://sui-testnet-endpoint.blockvision.org',
-    'https://rpc.ankr.com/sui_testnet',
-    'https://sui.blockpi.network/v1/rpc/public'
+    'https://sui-testnet-endpoint.blockvision.org'
 ];
--- a/portal/common/lib/constants.ts
+++ b/portal/common/lib/constants.ts
@@ -16,9 +16,9 @@ export const FALLBACK_PORTAL = "blob.store";
 export const RESOURCE_PATH_MOVE_TYPE = SITE_PACKAGE + "::site::ResourcePath";
 
 export const TESTNET_RPC_LIST = [
-    'https://fullnode.testnet.sui.io:443',
-    'https://sui-testnet.public.blastapi.io/',
-    'https://sui-testnet-endpoint.blockvision.org/',
+    'https://fullnode.testnet.sui.io',
+    'https://sui-testnet.public.blastapi.io',
+    'https://sui-testnet-endpoint.blockvision.org',
     'https://rpc.ankr.com/sui_testnet',
     'https://sui.blockpi.network/v1/rpc/public'
 ];
--- a/portal/common/lib/constants.ts
+++ b/portal/common/lib/constants.ts
@@ -15,14 +15,10 @@ export const FALLBACK_PORTAL = "blob.store";
 // The string representing the ResourcePath struct in the walrus_site package.
 export const RESOURCE_PATH_MOVE_TYPE = SITE_PACKAGE + "::site::ResourcePath";
 
-export const testnetRPCUrls = [
+export const TESTNET_RPC_LIST = [
     'https://fullnode.testnet.sui.io:443',
     'https://sui-testnet.public.blastapi.io/',
     'https://sui-testnet-endpoint.blockvision.org/',
     'https://rpc.ankr.com/sui_testnet',
-];
-
-export const prodRPCUrls = [
-    'https://sui-mainnet-endpoint.blockvision.org/',
-    'https://sui-mainnet.public.blastapi.io/'
+    'https://sui.blockpi.network/v1/rpc/public'
 ];
--- a/portal/common/lib/resource.test.ts
+++ b/portal/common/lib/resource.test.ts
@@ -7,9 +7,11 @@ import { fetchResource } from "./resource";
 import { HttpStatusCodes } from "./http/http_status_codes";
 import { checkRedirect } from "./redirects";
 import { fromBase64 } from "@mysten/bcs";
+import rpcSelectorSingleton from "./rpc_selector";
+import { SuiObjectResponse } from "@mysten/sui/client";
 
 // Mock SuiClient methods
-const multiGetObjects = vi.fn();
+const multiGetObjects = vi.spyOn(rpcSelectorSingleton, 'multiGetObjects');
 
 vi.mock("@mysten/sui/utils", () => ({
     deriveDynamicFieldID: vi.fn(() => "0xdynamicFieldId"),
@@ -59,6 +61,7 @@ describe("fetchResource", () => {
     });
 
     test("should fetch resource without redirect", async () => {
+
         // Mock object response
         multiGetObjects.mockResolvedValueOnce([
             {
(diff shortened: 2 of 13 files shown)


portal/common/lib/constants.ts
```diff
@@ -14,3 +14,15 @@ export const SITE_NAMES: { [key: string]: string } = {
...
+    'https://fullnode.testnet.sui.io:443',
+    'https://sui-testnet.public.blastapi.io/',
+    'https://sui-testnet-endpoint.blockvision.org/',
+    'https://rpc.ankr.com/sui_testnet',
+];
+
+export const prodRPCUrls = [
+    'https://sui-mainnet-endpoint.blockvision.org/',
+    'https://sui-mainnet.public.blastapi.io/'
+];
```
Tzal3x, line 28: Is this the complete and final list we want to include? 

`prodRPCUrls` should be skipped for now.

portal/common/lib/resource.test.ts
```diff
@@ -4,16 +4,15 @@
...
 import { fromBase64 } from "@mysten/bcs";
 
 // Mock SuiClient methods
 const multiGetObjects = vi.fn();
-const mockClient = {
-    multiGetObjects,
-} as unknown as SuiClient;
+// const mockClient = {
+//     multiGetObjects,
+// } as unknown as SuiClient;
```
Tzal3x, line 15: Let's remove commented-out code.

portal/common/lib/rpc_selector.ts
```diff
@@ -0,0 +1,169 @@
...
+            input,
+        ]);
+    }
+
+    public async call<T>(method: string, args: any[]): Promise<T> {
+        return this.invokeWithFailover<T>(method, args);
+    }
+}
+
+const rpcSelectorInstance = RPCSelector.getInstance();
```
Tzal3x, line 168: Nit: Rename this to `rpcSelectorSingleton`.

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -73,23 +72,19 @@ self.addEventListener("fetch", async (event) => {
...
-                : cachedResponse;
+            return cachedResponse;
         };
 
         // Fetch directly and fallback if necessary
         const fetchDirectlyOrProxy = async (): Promise<Response> => {
-            const response = await resolveAndFetchPage(parsedUrl, null);
-            return response.status === HttpStatusCodes.NOT_FOUND ? proxyFetch() : response;
+            const response = await resolveAndFetchPage(parsedUrl);
+            return response;
```
Tzal3x, line 87: I think we should still support the *devnet fallback*.

portal/common/lib/rpc_selector.ts
Tzal3x, line 1: What would happen in this scenario? 

- Fetch request for a `devnet` site (deployed with the old contract, destined to be handled by our **fallback portal**).
- The RPC returns 404. 
- In this case do we retry with another RPC fallback or do we propagate to the devnet fallback portal?

Tzal3x (changes requested):
Thanks @chariskms ! Left some minor comments. 🙇‍♂️

portal/common/lib/constants.ts
chariskms: These are all the nodes listed here (https://sui.io/developers#dev-tools) and are free to access.

```
export const TESTNET_RPC_LIST = [
    'https://fullnode.testnet.sui.io:443',
    'https://sui-testnet.public.blastapi.io/',
    'https://sui-testnet-endpoint.blockvision.org/',
    'https://rpc.ankr.com/sui_testnet',
    'https://sui.blockpi.network/v1/rpc/public'
];
```

portal/common/lib/constants.ts
```diff
@@ -14,3 +14,11 @@ export const SITE_NAMES: { [key: string]: string } = {
 export const FALLBACK_PORTAL = "blob.store";
 // The string representing the ResourcePath struct in the walrus_site package.
 export const RESOURCE_PATH_MOVE_TYPE = SITE_PACKAGE + "::site::ResourcePath";
+
+export const TESTNET_RPC_LIST = [
+    'https://fullnode.testnet.sui.io:443',
+    'https://sui-testnet.public.blastapi.io/',
+    'https://sui-testnet-endpoint.blockvision.org/',
+    'https://rpc.ankr.com/sui_testnet',
+    'https://sui.blockpi.network/v1/rpc/public'
```
giac-mysten, line 23: Nit: use a common format for the urls. I.e., all have/don't have the trailing `/`.

portal/common/lib/constants.ts
```diff
@@ -14,3 +14,11 @@ export const SITE_NAMES: { [key: string]: string } = {
 export const FALLBACK_PORTAL = "blob.store";
 // The string representing the ResourcePath struct in the walrus_site package.
 export const RESOURCE_PATH_MOVE_TYPE = SITE_PACKAGE + "::site::ResourcePath";
+
+export const TESTNET_RPC_LIST = [
+    'https://fullnode.testnet.sui.io:443',
```
giac-mysten, line 19: Nit: don't think the port number is necessary here (`https` is `443`)

portal/common/lib/resource.test.ts
giac-mysten, line 1: For the future: can we separate tests into their own directory? 🙏

portal/common/lib/rpc_selector.ts
```diff
@@ -0,0 +1,159 @@
...
+        ]);
+    }
+
+    public async call<T>(method: string, args: any[]): Promise<T> {
+        return this.invokeWithFailover<T>(method, args);
+    }
+}
+
+const rpcSelectorSingleton = RPCSelector.getInstance();
+export default rpcSelectorSingleton;
```
giac-mysten, line 159: Remark: in general, i'm not a huge fan of singletons, and prefer dependency injection whenever possible. Yet, this is good enough for now imo.

portal/common/lib/rpc_selector.ts
```diff
@@ -0,0 +1,159 @@
...
+    // General method to call clients and return the first successful response.
+    private async invokeWithFailover<T>(methodName: string, args: any[]): Promise<T> {
+        if (this.clients.length === 0) {
+            throw new Error("No available clients to handle the request.");
+        }
+
+        const isNoSelectedClient = !this.selectedClient;
+        if (isNoSelectedClient) {
+            return await this.callFallbackClients<T>(methodName, args);
+        }
```
giac-mysten, line 50: nit: this check could be removed i think, since the following try block would result in `callFallbackClients` if there is no selected client

portal/common/lib/rpc_selector.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    // Attempt to call the method on the selected client with a timeout.
+    private async callSelectedClient<T>(methodName: string, args: any[]): Promise<T> {
+        const method = (this.selectedClient as any)[methodName] as Function;
+        if (!method) {
+            throw new Error(`Method ${methodName} not found on selected client`);
+        }
+        const timeoutDuration = 5000;
```
giac-mysten, line 67: Remark: this is a bit arbitrary and may be too low, especially if the call needs to load a resource. 

Question: Is there a way to set a timeout on connection establishment, instead of on the whole call?

portal/common/lib/rpc_selector.ts
```diff
@@ -0,0 +1,159 @@
...
+    // Fallback to querying all clients using Promise.any.
+    private async callFallbackClients<T>(methodName: string, args: any[]): Promise<T> {
+        const clientPromises = this.clients.map((client) =>
+            new Promise<{ result: T; client: SuiClient }>(async (resolve, reject) => {
+                try {
+                    const method = (client as any)[methodName] as Function;
+                    if (!method) {
+                        reject(new Error(`Method ${methodName} not found on client`));
+                        return;
+                    }
```
giac-mysten, line 94: Question: why should this occur?

portal/common/lib/rpc_selector.ts
```diff
@@ -0,0 +1,159 @@
...
+                    reject(error);
+                }
+            }),
+        );
+
+        try {
+            const { result, client } = await Promise.any(clientPromises);
+            // Update the selected client for future calls.
+            this.selectedClient = client;
+            return result;
```
giac-mysten, line 111: Praise: I like this! 🚀 

Question: what is the "lifetime" of a selected client, depending on the way the portal is deployed? Something like:
- service worker: until the service worker is refreshed
- server side on edge functions: every call has a different one?

Suggestion: log which client was selected, for easier testing?

giac-mysten (approved):
Thanks so much @chariskms , this already looks great. Left a few comments and questions. 

In general, did you try this with some "failed" full node? does the switching work well when a failure occurs? Is the latency noticeable?

portal/common/lib/rpc_selector.ts
chariskms: Since the first use of the fallback function is considered default behavior, I’d suggest keeping the if statement, as throwing an exception would likely be more costly for the standard flow.

portal/common/lib/resource.test.ts
chariskms: Yes, that would be great to have.

portal/common/lib/rpc_selector.ts
chariskms: It might be a bit challenging since the actual call happens through the Sui client, but I can extend the time.

portal/common/lib/rpc_selector.ts
chariskms: We use this approach to avoid duplicating code across each Sui client method, especially since we already have a generic call function for cases like Suins.

Tzal3x (approved):
Sweet, let's test it on testnet! 🤞

PR #282, opened 2024-11-01
refactor: check default directories

giac-mysten:
minor refactor.

- uses the already existing default selection function
- simplifies the selection of paths.

### DIFF
--- a/site-builder/src/main.rs
+++ b/site-builder/src/main.rs
@@ -11,13 +11,14 @@ mod util;
 mod walrus;
 use std::path::PathBuf;
 
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Result};
 use clap::{Parser, Subcommand};
 use futures::TryFutureExt;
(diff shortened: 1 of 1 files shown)


site-builder/src/main.rs
```diff
@@ -206,73 +209,40 @@ mod default {
...
-    let home_dir = match home::home_dir() {
-        Some(dir) => dir,
-        None => {
-            return Err(anyhow::anyhow!(
-                "failed to get home directory. $HOME needs to be defined."
-            ));
-        }
+/// Returns the default paths for the sites-config.yaml file.
+pub fn sites_config_default_paths() -> Vec<PathBuf> {
+    let mut default_paths = vec![SITES_CONFIG_NAME.into()];
```
Tzal3x, line 214: praise: nice idea to add every path to the same vec (including the default argument of the config)

Tzal3x (approved):

PR #284, opened 2024-11-01
feat(site-builder): add option to limit parallel blob ID computation

giac-mysten:
Adds a `--max-concurrent <LIMIT>` cli flag to the `publish` and `update` commands. If specified, it limits the number of concurrent calls to the Walrus CLI to compute the blob ID.

This is useful to reduce the memory footprint of the site builder, at the cost of a longer time spent computing blob IDs.

### DIFF
--- a/site-builder/src/publish.rs
+++ b/site-builder/src/publish.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
+    num::NonZeroUsize,
     path::{Path, PathBuf},
     sync::mpsc::channel,
 };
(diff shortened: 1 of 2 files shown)


site-builder/src/publish.rs
```diff
@@ -57,6 +58,9 @@ pub struct PublishOptions {
     /// See the `list-directory` command. Warning: Rewrites all `index.html` files.
     #[clap(long, action)]
     pub list_directory: bool,
+    /// The maximum number of concurrent calls to the Walrus CLI for the computation of blob IDs.
+    #[clap(long)]
+    max_concurrent: Option<NonZeroUsize>,
```
Tzal3x, line 63: Should we add a default? Whenever someone needs to set up the `max_concurrent`, most likely they will have run into an error.

Tzal3x (approved):
Nice 🚀

PR #288, opened 2024-11-06
fix(site-builder): fix issue when only deleting resources

giac-mysten:


Tzal3x (approved):
Thanks for the fix @giac-mysten 🙇

PR #289, opened 2024-11-06
feat(site-builder): use multiple ptbs to upload large sites

giac-mysten:
The site builder now uses multiple ptbs in case the site update requires more than 200 resource operations.


Tzal3x (approved):
Nice 💯

PR #290, opened 2024-11-08
Add new feature : upload file to walrus by multiple workers

v1xingyue:
Mainly changed:

1. Add config params `workers` with `Config` 
2. Add clap params `workers` with `GeneralArgs`
3. upload and publish use multiple workers as configuration. 

Example usage: 

```shell
site-builder --workers 10 --config ./client_config.yaml update --epochs 200 ./site
```

### DIFF
--- a/site-builder/src/main.rs
+++ b/site-builder/src/main.rs
@@ -75,11 +75,15 @@ pub(crate) struct GeneralArgs {
     #[clap(short, long)]
     #[serde(default = "default::gas_budget")]
     gas_budget: Option<u64>,
+    /// The number of workers to use for parallel processing.
+    #[clap(long, default_value_t = 1)]
+    workers: usize,
 }
 
 impl Default for GeneralArgs {
     fn default() -> Self {
         Self {
+            workers: 1,
             rpc_url: None,
             wallet: None,
             walrus_binary: default::walrus_binary(),
@@ -171,6 +175,8 @@ pub(crate) struct Config {
     pub package: ObjectID,
     #[serde(default)]
     pub general: GeneralArgs,
+    #[serde(default)]
+    pub workers: usize,
 }
 
 impl Config {
@@ -179,6 +185,11 @@ impl Config {
     /// The values in `other_general` take precedence.
     pub fn merge(&mut self, other_general: &GeneralArgs) {
         self.general.merge(other_general);
+        if self.workers == 0 {
+            if other_general.workers > 0 {
+                self.workers = other_general.workers;
+            }
+        }
     }
 
     pub fn walrus_binary(&self) -> String {
--- a/site-builder/src/site/manager.rs
+++ b/site-builder/src/site/manager.rs
@@ -107,10 +107,9 @@ impl SiteManager {
 
     /// Publishes the resources to Walrus using parallel uploads.
     async fn publish_to_walrus<'b>(&mut self, updates: &[&ResourceOp<'b>]) -> Result<()> {
-        // 创建信号量来控制并发数
+        // create semaphore to control concurrency
         let semaphore = Arc::new(Semaphore::new(self.config.workers));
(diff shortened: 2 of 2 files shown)


v1xingyue:
> Thank you for the contribution @v1xingyue ! We will review your PR shortly.

Ya.  But  I'm not sure what happened. 
Sometimes error will happend like this :

```
RPC call failed: ErrorObject { code: ServerError(-32002), message: "Failed to sign transaction by a quorum of validators because one or more of its objects is reserved for another transaction. Retried transaction 272PdQfqLqJSZAT42nLjePRji5CXF967AnBPpJCrdpFW (succeeded) because it was able to gather the necessary votes. Other transactions locking these objects:\n- 272PdQfqLqJSZAT42nLjePRji5CXF967AnBPpJCrdpFW (stake 75.92)
```

I don't know why this happened and what should I do?

Tzal3x (commented):
To me it seems ok, and it works on my machine. Nice work! 

@giac-mysten since you are more familiar with Rust would you mind reviewing also the use of the semaphore?

giac-mysten:
> > Thank you for the contribution @v1xingyue ! We will review your PR shortly.
> 
> Ya. But I'm not sure what happened. Sometimes error will happend like this :
> 
> ```
> RPC call failed: ErrorObject { code: ServerError(-32002), message: "Failed to sign transaction by a quorum of validators because one or more of its objects is reserved for another transaction. Retried transaction 272PdQfqLqJSZAT42nLjePRji5CXF967AnBPpJCrdpFW (succeeded) because it was able to gather the necessary votes. Other transactions locking these objects:\n- 272PdQfqLqJSZAT42nLjePRji5CXF967AnBPpJCrdpFW (stake 75.92)
> ```
> 
> I don't know why this happened and what should I do?

Hey @v1xingyue , many thanks for the contribution. Unfortunately parallel uploads are quite tricky. The problem is that the `walrus` clients all use the same wallet, and therefore it may happen that they try to use the same gas coin or storage object at the same time, creating a race condition that is not allowed on chain (it would be basically double spending!)

I'm working on the walrus core side to improve this functionality on the `walrus` client side. However, until that is ready, I think there is no good way to implement parallel uploads. 

Please reopen the pull request if I misunderstoon anything! And thanks again for working with walrus sites.

v1xingyue:
> > > Thank you for the contribution @v1xingyue ! We will review your PR shortly.
> > 
> > 
> > Ya. But I'm not sure what happened. Sometimes error will happend like this :
> > ```
> > RPC call failed: ErrorObject { code: ServerError(-32002), message: "Failed to sign transaction by a quorum of validators because one or more of its objects is reserved for another transaction. Retried transaction 272PdQfqLqJSZAT42nLjePRji5CXF967AnBPpJCrdpFW (succeeded) because it was able to gather the necessary votes. Other transactions locking these objects:\n- 272PdQfqLqJSZAT42nLjePRji5CXF967AnBPpJCrdpFW (stake 75.92)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't know why this happened and what should I do?
> 
> Hey @v1xingyue , many thanks for the contribution. Unfortunately parallel uploads are quite tricky. The problem is that the `walrus` clients all use the same wallet, and therefore it may happen that they try to use the same gas coin or storage object at the same time, creating a race condition that is not allowed on chain (it would be basically double spending!)
> 
> I'm working on the walrus core side to improve this functionality on the `walrus` client side. However, until that is ready, I think there is no good way to implement parallel uploads.
> 
> Please reopen the pull request if I misunderstoon anything! And thanks again for working with walrus sites.

Seems, I got the problem. 
What if I split coin first, and each transaction use the diffrent coin as the split usage. 
Then that will be fine.

PR #292, opened 2024-11-08
feat(portal): remove devnet fallback

chariskms:
The portal previously included a mechanism to redirect requests to devnet as a fallback in the event of a failure or a 404 (Not Found) response. **This pull request removes that fallback functionality**.

**Changes** 

- Removed the fallback to FALLBACK_DEVNET_PORTAL in the main routing logic.
- Simplified the fetch process by eliminating proxyFetch and fallback handling.
- Streamlined code to directly return responses or appropriate errors without devnet fallback references.

### DIFF
--- a/portal/server/app/route.ts
+++ b/portal/server/app/route.ts
@@ -31,28 +31,11 @@ export async function GET(req: Request) {
     const requestDomain = getDomain(url);
 
     if (requestDomain == portalDomain && parsedUrl && parsedUrl.subdomain) {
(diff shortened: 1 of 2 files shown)


Tzal3x (changes requested):
Nice, needs 2 things and it will be ready. 🙇‍♂️

Tzal3x (changes requested):

Tzal3x (approved):
Awesome 💯

PR #293, opened 2024-11-11
feat(portal): update portal domain name resolution using length

Tzal3x:
- Add a new `portalNameLength` wherever it applies. 
- Update the `domain_parsing.tests` to cover the new functionality.

The PR is pointing to the `testnet` branch as it is the best branch to test this. 
It uses a `sw-tnet.blocksite.net` domain + the environment variables of the portal name length have already been set. 

Note: In the future, if we make the common library more object oriented, this awkward argument passing can be avoided.

### DIFF
--- a/portal/common/lib/domain_parsing.test.ts
+++ b/portal/common/lib/domain_parsing.test.ts
@@ -28,6 +28,20 @@ describe('getDomain', () => {
     })
 })
 
+const getDomainWithPortalNameLengthTestCases: [string, string][] = [
+    ['https://sw-tnet.blocksite.net', 'sw-tnet.blocksite.net'],
+    ['https://subname.sw-tnet.blocksite.net', 'sw-tnet.blocksite.net']
+]
+
+describe('getDomain with portal name length', () => {
+    getDomainWithPortalNameLengthTestCases.forEach(([input, expected]) => {
+        test(`${input} -> ${expected}`, () => {
+            const domain = getDomain(new URL(input), 21)
+                expect(domain).toEqual(expected)
+        })
+    })
+})
+
 const getSubdomainAndPathTestCases: [string, DomainDetails][] = [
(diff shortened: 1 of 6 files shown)


portal/common/lib/domain_parsing.test.ts
```diff
@@ -5,6 +5,8 @@ import { describe, expect, test } from 'vitest'
 import { getDomain, getSubdomainAndPath } from './domain_parsing'
 import { DomainDetails } from './types'
 
+const PORTAL_DOMAIN_NAME_LENGTH = 21
```
giac-mysten, line 8: question: what other alternatives do we have to avoid this rather ambiguous length? Specifying the full portal name? or the number of splits?

portal/common/lib/domain_parsing.ts
```diff
@@ -29,13 +29,18 @@ export function getSubdomainAndPath(url: URL): DomainDetails | null {
     {domain: name.walrus.site,
     { subdomain: "subname", path: "/index.html"}}
 */
-function splitUrl(url: URL): UrlExtract {
+function splitUrl(url: URL, portalNameLength?: Number): UrlExtract {
```
giac-mysten, line 32: Question: does this work when developing on localhost?

giac-mysten (approved):
Thanks @Tzal3x , this works. Not sure specifying only the length is the best way to go though, looks like it could break easily. Up to you though.

portal/common/lib/domain_parsing.ts
Tzal3x: Yes! Tests reassure us too! 👍

portal/common/lib/domain_parsing.test.ts
Tzal3x: I think the next right approach would be to specify the portal name, but this would also require to set on every vercel deployment the corresponding portal domain explicitly. 

Since all the subdomains of blocksite have the same length, this is an easy way of going around this.

PR #296, opened 2024-11-12
fix(portal): sw portal not loading external resources

giac-mysten:


Tzal3x (approved):
Thank you @giac-mysten for the quick fix!

PR #297, opened 2024-11-12
fix(portal): rpcselector generic call

chariskms:
Fix RPCSelector Generic Call and Logging Enhancements

**Changes:**

- Generic Call Fix: Corrected call method syntax in RPCSelector.
- Logging: Added logs for fallback and client selection tracking.
- Test Update: Aligned resolveSuiNsAddress tests with new call method structure.


Tzal3x (approved):
🔥

PR #298, opened 2024-11-13
feat: create workflow for tag creation

Tzal3x:
- [x] Create a `tag.yml` workflow that creates a tag given a commit and a tag name.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,0 +1,51 @@
+# Tags a specified commit with a given tag name.
+# Triggers: Manually via workflow_dispatch, Automatically on repository_dispatch.
+
+name: Tag Site Builder Branch
+
+on:
+    repository_dispatch:
+        types: [tag-walrus-sites-branch]
+
+    workflow_call:
+        inputs:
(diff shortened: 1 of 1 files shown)


ebmifa (approved):

giac-mysten:
do we have a use case for this workflow yet? what is triggering it?

giac-mysten:
And regarding the use case? 
I think @ebmifa is looking into the binary workflow and versioning for sites (see [linear](https://linear.app/mysten-labs/issue/PRO-232/set-up-ci-workflows-for-site-builder-binaries-from-walrus-sites)). Probably good to wait for the outcome of that flow?

Tzal3x:
> And regarding the use case? I think @ebmifa is looking into the binary workflow and versioning for sites (see [linear](https://linear.app/mysten-labs/issue/PRO-232/set-up-ci-workflows-for-site-builder-binaries-from-walrus-sites)). Probably good to wait for the outcome of that flow?

Sounds good, I just subscribed to the issue! Changing the PR to draft until further notice.

mlegner:
I think this can be closed since #308 was merged.

PR #300, opened 2024-11-13
refactor: create logger class abstraction

Tzal3x:
This class aims to be runtime-agnostic,
meaning that it should be able to run
in both the browser and on a server.

It is useful for cases where someone
needs to use a logging mechanism that
might be only supported on one runtime.

For example [amplitude](https://amplitude.com/docs/sdks/sdk-quickstart) provides different
interfaces (sdks) for different runtimes.

### DIFF
--- a/portal/common/lib/logger.ts
+++ b/portal/common/lib/logger.ts
@@ -1,0 +1,57 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// Provides a simple logger interface function for
+/// logging messages on different runtimes.
+type LoggingPredicate = (...args: any) => void;
+
+/**
+ * Logger used as an abstraction for logging messages on different runtimes.
+ */
+ class Logger {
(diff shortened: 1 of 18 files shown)


portal/common/lib/logger.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/// Provides a simple logger interface function for
+/// logging messages on different runtimes.
+type LoggingPredicate = (...args: any) => void;
+
+/**
+ * Logger used as an abstraction for logging messages on different runtimes.
+ */
+ class Logger {
+    private infoPredicate: LoggingPredicate;
```
giac-mysten, line 12: Question: is there no level below info? i'd assume something like `debug` should exist?

giac-mysten (approved):
LGTM!

portal/common/lib/logger.ts
Tzal3x: Created a new PR to avoid conflicts with the other branches https://github.com/MystenLabs/walrus-sites/pull/307

PR #301, opened 2024-11-14
fix(service-worker & cache): remove checkCachedVersionMatchesOnChain

Tzal3x:
1. Fixed the issue of the caching mechanism always "missing" in the service worker.
2. Updated the CACHE_EXPIRATION_TIME to 1 hour.

The resolveObjectId function would only be called on the site::Site object and not for each Resource (which makes sense as it is doing the suins-to-objectId resolution).

However, this meant that a lot of requests would be tagged as "cache miss" incorrectly, because the
site::Site version would be compared with the
site::Resource version inside the checkCachedVersionMatchesOnChain.

Thinking more about it, the end goal of using a cache is to avoid making requests to the RPC nodes.
Always checking for the sui object version contradicts this. Therefore I decided to remove the checkCachedVersionMatchesOnChain altogether.

Remember that caches expire, so at some point the resource contents will be updated anyway.

### DIFF
--- a/portal/worker/src/caching.ts
+++ b/portal/worker/src/caching.ts
@@ -6,36 +6,24 @@ import { DomainDetails } from "@lib/types";
 import rpcSelectorSingleton from "@lib/rpc_selector";
 
 const CACHE_NAME = "walrus-sites-cache";
-const CACHE_EXPIRATION_TIME = 24 * 60 * 60 * 1000; // 24 hours in milliseconds
+const CACHE_EXPIRATION_TIME = 60 * 60 * 1000; // 1 hour in milliseconds
 
 /**
  * Respond to the request using the cache API.
  */
 export default async function resolveWithCache(
-    resolvedObjectId: string,
     parsedUrl: DomainDetails,
     urlString: string,
 ): Promise<Response> {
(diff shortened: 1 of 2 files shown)


giac-mysten:
Thanks @Tzal3x for this improvement. While I agree with the general direction, there are several point i'm not sure about.

First, a blanket 1h expiration for the cache may be way too high for certain assets, and way too low for others (especially for SuiNS names and RPC requests).  Now that we have proper header support, maybe the browser will be able to do most of the caching, so this is not really a problem (_if_ the browser cache hits before the request is sent to the service worker). One possible solution would be to use the caching header to set the expiration on the sw cache.

Another alternative is to have a short cache (10s to 1 min) for rpc requests. This would really massively reduce the amount of requests, while still allowing fast updates.

Tzal3x:
> Another alternative is to have a short cache (10s to 1 min) for rpc requests. This would really massively reduce the amount of requests, while still allowing fast updates.

We could push this fix for now, with the `1 min` caching. Later we can re-iterate and find a way to also use cache response headers.

giac-mysten (approved):
This works, thanks. Can you open an issue to track the improvements to caching?

PR #302, opened 2024-11-14
feat: integrate sentry to server portal

Tzal3x:
**Changes**
- Integrate Sentry to the Logger implementation on the server portal.
- Update logs to send object looking like`{message: String, other-info: string}`.

For staging, I created a walrus-sites-sp-testnet project on Sentry which you can find [here](https://mysten-labs.sentry.io/traces/?environment=development&environment=vercel-preview&project=4508298139271168&statsPeriod=24h).

Note: the logs it contains are also from local development and there are a lot of trash from previous experiments. Need to clear it up later when I get access. 

**Future work**: 
- [x] Enrich logs in the common library. Maybe rephrase the logs too.

### DIFF
--- a/portal/server/app/route.ts
+++ b/portal/server/app/route.ts
@@ -26,8 +26,12 @@ logger.setWarnPredicate(args => {
 } );
 logger.setInfoPredicate(args => {
     addLoggingArgsToSentry(args);
-    Sentry.addBreadcrumb({ message: args.message, data: args })
+    Sentry.addBreadcrumb({ message: args.message, data: args, level: 'info'})
 } );
+logger.setDebugPredicate(args => {
+    addLoggingArgsToSentry(args);
+    Sentry.addBreadcrumb({ message: args.message, data: args, level: 'debug' })
+});
 
 export async function GET(req: Request) {
     const originalUrl = req.headers.get("x-original-url");
--- a/portal/server/app/route.ts
+++ b/portal/server/app/route.ts
@@ -8,9 +8,26 @@ import { resolveAndFetchPage } from "@lib/page_fetching";
 import logger from "@lib/logger";
 import * as Sentry from "@sentry/node";
 
-logger.setErrorPredicate((...args: any) => { Sentry.captureException(new Error(...args)) } );
-logger.setWarnPredicate((...args: any) => { Sentry.addBreadcrumb({ message: 'Warning', data: [...args] }) } );
-logger.setInfoPredicate((...args: any) => { Sentry.addBreadcrumb({ message: 'Info', data: [...args] }) } );
+function addLoggingArgsToSentry(args: { [key: string]: any }) {
+    Object.entries(args).forEach(([key, value]) => {
+        if (key !== "message") { // Skipping the 'message' key
+            console.log(`${key}: ${value}`)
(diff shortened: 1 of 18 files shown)


portal/server/app/route.ts
```diff
@@ -5,6 +5,29 @@ import { getDomain, getSubdomainAndPath } from "@lib/domain_parsing";
...
+    Sentry.captureException(new Error(args.message ))
+});
+logger.setWarnPredicate(args => {
+    addLoggingArgsToSentry(args);
+    Sentry.addBreadcrumb({ message: args.message, data: args })
+} );
+logger.setInfoPredicate(args => {
+    addLoggingArgsToSentry(args);
+    Sentry.addBreadcrumb({ message: args.message, data: args })
+} );
```
Tzal3x, line 30: This is the part that the logger predicates are defined for the server portal, so that it utilizes Sentry. 
The `breadcrumbs` are additional information about a trace (i.e. custom logs). 
The `tags` are added to each event so we can query the logs on the Sentry UI.

portal/server/next.config.mjs
```diff
@@ -16,4 +20,37 @@ const nextConfig = {
...
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/
+
+// Upload a larger set of source maps for prettier stack traces (increases build time)
+widenClientFileUpload: true,
+
+// Route browser requests to Sentry through a Next.js rewrite to circumvent ad-blockers.
+// This can increase your server load as well as your hosting bill.
+// Note: Check that the configured route will not match with your Next.js middleware, otherwise reporting of client-
+// side errors will fail.
+tunnelRoute: "/monitoring",
```
Tzal3x, line 43: Might need to remove the `tunnelRoute`.

portal/server/app/route.ts
giac-mysten: question: isn't the information duplicated if the  tags and the breadcrumbs both contain `args`?  is this useful?

portal/server/next.config.mjs
giac-mysten: question: why remove?

portal/server/instrumentation.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export async function register() {
+  if (process.env.NEXT_RUNTIME === 'nodejs') {
+    await import('./sentry.server.config');
+  }
+
+  if (process.env.NEXT_RUNTIME === 'edge') {
+    await import('./sentry.edge.config');
+  }
+}
```
giac-mysten, line 14: question: do we need both sentry edge and server?

portal/server/next.config.mjs
Tzal3x: Because it might break the routing of the portal. Users might not be able to define a `/monitoring` route in their walrus site.

Tzal3x: Scratch that, I tried it out and there was no conflict with the WS, meaning that the portal will look up for a /monitoring site resource.

portal/server/instrumentation.ts
Tzal3x: For local development as far as I understand we need nodejs, while edge for vercel deployments.

PR #307, opened 2024-11-20
feat(common): add a new -debug- logging level

Tzal3x:
- Create a new debug level, used as the highest level of logging.
- Add more descriptive comments to the logger class.
- Update the logging level of some of the logs used in the server portal.


Tzal3x:
@giac-mysten feel free to replace the log levels using `logger.debug` instead of `logger.info` wherever you see fit.

giac-mysten (approved):
Thanks! very nice to have. Re: the downgrading to debug, I think all prints that are simple sanity checks (ie, object id conversions, returned object contents, etc) should be debug. Can be done in a followup

PR #309, opened 2024-11-21
add remove_header operate

v1xingyue:


Tzal3x:
Hey, why would we want to remove the function of removing headers?

PR #315, opened 2024-12-04
feat: use type origin map to get correct struct types after upgrades

karlwuest:


giac-mysten (approved):
This is really cool! thanks for the contribution @karlwuest

PR #316, opened 2024-12-04
feat: add second rpc that support suins

chariskms:
Add `https://testnet.suiet.app` as the second rpc that supports suins on testnet.


Tzal3x (approved):
Awesome, let's try this out!

PR #317, opened 2024-12-05
fix: update info-output format and make request for n_shards more robust

mlegner:
Updates the info-output format to align it with changes in https://github.com/MystenLabs/walrus/pull/1204.


giac-mysten (approved):
thanks for the quick fix @mlegner ! the partial deserialization is quite cool :)

PR #318, opened 2024-12-05
feat(site-builder): add new package to config, expose destroy command

giac-mysten:
Closes SEINT-357
Closes SEINT-371

### DIFF
--- a/move/walrus_site/sources/site.move
+++ b/move/walrus_site/sources/site.move
@@ -204,11 +204,11 @@ module walrus_site::site {
     }
 
     /// Deletes a site object.
-    /// WARNING: This function does **NOT** delete the dynamic fields!
-    /// Make sure to call this function after deleting manually
-    /// all the dynamic fields attached to the sites object.
-    /// If you don't delete the dynamic fields, they will become
-    /// unaccessible and you will not be able to delete them in the future.
+    ///
+    /// NB: This function does **NOT** delete the dynamic fields! Make sure to call this function
+    /// after deleting manually all the dynamic fields attached to the sites object. If you don't
+    /// delete the dynamic fields, they will become unaccessible and you will not be able to delete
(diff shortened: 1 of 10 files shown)


Tzal3x (commented):
Sweet! Thank you 🙇‍♂️ Some newborn questions: 

1. This only deletes already "expired" sites right? Sites that the blob of the resources have been gone past their epoch expiration, so they are deleted. And the user just wants to take the reimbursement by deleting the no-longer needed sui objects. 
2. If a user needs to delete a site, before that expiring, this functionality is not supported yet in the `site-builder` right? Meaning that during publishing he should define the resource blobs as deleteable, delete them, and then have a separate function (maybe different than destroy?) that deletes the blobs first, and then calling `site-builder destroy`?

giac-mysten:
>     1. This only deletes already "expired" sites right? Sites that the blob of the resources have been gone past their epoch expiration, so they are deleted. And the user just wants to take the reimbursement by deleting the no-longer needed sui objects.

>     2. If a user needs to delete a site, before that expiring, this functionality is not supported yet in the `site-builder` right? Meaning that during publishing he should define the resource blobs as deleteable, delete them, and then have a separate function (maybe different than destroy?) that deletes the blobs first, and then calling `site-builder destroy`?

Ah this is actually an excellent point. Currently, the destroy command does not touch the blobs on walrus. Only removes the Site object and dynamic fields from Sui. Which may not be what people want. However, we do not support deletable sites at the moment. So a bit more thinking may be actually needed here to decide what is best to do.

Possibly a good idea would be to always publish sites as deletable blobs, unless someone explicitly adds the `--permanent` flag.

Tzal3x (approved):
Ok, let's merge this and we then can proceed by building the deletable blobs functionality in a following PR.

PR #323, opened 2024-12-10
refactor(portal): make sp locally deployable and vercel independent

Tzal3x:
- Create `sentry_logger.ts`
- Create `custom_blocklist_checker.ts`
- Only integrate sentry and edge config upon production deployments, i.e. when running `pnpm run serve:prod:server`


Tzal3x:
> Nice one! Do we have documentation already on how to deploy locally vs on vercel?

I think we don't have yet, but we already have a ticket for this so it won't be skipped! ✍️

PR #324, opened 2024-12-11
feat(portal): SEINT-406 hide full node urls from the server portal code

Tzal3x:
Note: This requires updating our environment variables on Vercel! 
- Specify hard coded env variables for the service worker.
- Create a `common/.env.test` for simulating env vars for testing. 
- Add `.env.example`


portal/common/lib/rpc_selector.ts
```diff
@@ -44,12 +43,25 @@ class RPCSelector implements RPCSelectorInterface {
         this.selectedClient = undefined;
     }
 
+    /**
+     * Get the list of RPC URLs from the environment variable.
+     */
+    private static parseRPCList(): string[] {
+        const rpcList = process.env.RPC_URL_LIST;
```
giac-mysten, line 50: Question: does this work even when running the portal locally?

portal/common/lib/rpc_selector.ts
```diff
@@ -44,12 +43,25 @@ class RPCSelector implements RPCSelectorInterface {
...
+        if (!rpcList) {
+            throw new Error("No RPC list found in environment variables");
+        }
+        return rpcList.split(",");
```
giac-mysten, line 54: Question: how is the list provided? as comma-separated i assume?

portal/common/lib/rpc_selector.ts
```diff
@@ -80,7 +92,10 @@ class RPCSelector implements RPCSelectorInterface {
         }
 
         const timeoutPromise = new Promise<never>((_, reject) =>
-            setTimeout(() => reject(new Error("Request timed out")), RPC_REQUEST_TIMEOUT_MS),
+            setTimeout(() => reject(
+                new Error("Request timed out")),
+                Number(process.env.RPC_REQUEST_TIMEOUT_MS) ?? 7000
```
giac-mysten, line 97: Question: why `7000`? Why should `process.env.RPC_REQUEST_TIMEOUT_MS` not be present?
Suggestion: make this a const at least.

giac-mysten (approved):
Thanks @Tzal3x , this looks good. 

One thing I noticed is that we have a proliferation of of env variables and files, and that we need to check if they are set everywhere. 

Is there a way we can give more structure to the configuration, and make sure it's well formed?

portal/common/lib/rpc_selector.ts
Tzal3x: Yes, as long as the environment variable is specified. We might need to provide a `.env.example` to our repo for people that want to do zero configuration.

portal/common/lib/rpc_selector.ts
Tzal3x: Yes the list is provided as comma separated. e.g.
```
RPC_LIST=https://node1.com,https://node2.com,https://node3.com
```

portal/common/lib/rpc_selector.ts
Tzal3x: That's a remnant actually. I think we initially decided it to be this big out of convenience. 
Not sure what the best value would be for this, but we can try different configurations to see what fits best.

PR #326, opened 2024-12-12
refactor(common-lib): use a more object-oriented-programming approach

Tzal3x:
# Summary

I noticed that we need more flexibility as more features are coming in. Closes SEINT-428.

Moving on to an object-oriented approach by defining some classes and composing new ones based on them using the composition pattern will help us achieve that. 

This way we can avoid spiraling down to multiple `if-else` clauses whenever a new feature request is coming.

In upcoming PRs, we can create a `PageFetcherFactory` class that generates a different `pageFetcher` which will contain different rpcURLs, domain resolution logic etc based on the type of site (based on the allowlist) that we will serve.

## Changelog
- Create a `PageFetcher` class.
- Create a `ResourceFetcher` class.
- Create a `WalrusSitesRouter` class.
- `rpcSelector` is no longer a singleton. A different`RPCSelector` can be used for: suin resolution, resource fetching and routes fetching without adding complex conditional flow control in the `common/` library's code.
- Fix environment variable parsing in the service worker. 
- Update tests to depict the new changes.


portal/common/lib/page_fetching.ts
```diff
@@ -18,166 +18,176 @@ import {
...
-        if (!isObjectId) {
-            return resolveObjectResult;
+* Includes all the logic for fetching a walrus site.
+*/
+export class PageFetcher {
+    constructor(
+        private resourceFetcher: ResourceFetcher,
+        private suinsResolver: SuiNSResolver,
+        private wsRouter: WalrusSitesRouter
+    ){}
```
giac-mysten, line 34: Question: it becomes apparent here that there is some ambiguous use of "page", "site", and "resource".
IIUC, we never fetch sites or pages. (this is all my fault, btw 😓 )

what about: 
- UrlFetcher: this represents the routing of a url to a resource
- ResourceFetcher: as it is now, resource->blob

portal/common/lib/page_fetching.ts
```diff
@@ -18,166 +18,176 @@ import {
...
-    resolvedObjectId: string | null,
-    blocklistChecker?: BlocklistChecker
-): Promise<Response> {
-    logger.debug({ message: "parsed-url", subdomain: parsedUrl.subdomain, path: parsedUrl.path });
-    if (!resolvedObjectId) {
-        const resolveObjectResult = await resolveObjectId(parsedUrl);
-        const isObjectId = typeof resolveObjectResult == "string";
-        if (!isObjectId) {
-            return resolveObjectResult;
+* Includes all the logic for fetching a walrus site.
```
giac-mysten, line 27: Suggestion: see the previous comment
```suggestion
* Includes all the logic for fetching a walrus URL.
```

portal/common/lib/page_fetching.ts
```diff
@@ -18,166 +18,176 @@ import {
...
+        private wsRouter: WalrusSitesRouter
+    ){}
+
+    /**
+     * Resolves the subdomain to an object ID, and gets the corresponding resources.
+     *
+     * The `resolvedObjectId` variable is the object ID of the site that was previously resolved. If
+     * `null`, the object ID is resolved again.
+     */
+    public async resolveAndFetchPage(
```
giac-mysten, line 42: As before, you get the idea
```suggestion
    public async resolveAndFetchUrl(
```

portal/server/app/route.ts
```diff
@@ -4,16 +4,32 @@
...
+if (!rpcUrlList) {
+    throw new Error("Missing RPC_URL_LIST environment variable");
+}
+const rpcSelector = new RPCSelector(rpcUrlList.split(','));
+const pageFetcher = new PageFetcher(
+    new ResourceFetcher(rpcSelector),
+    new SuiNSResolver(rpcSelector),
+    new WalrusSitesRouter(rpcSelector)
+);
+
```
giac-mysten, line 32: Praise: yesss I like this
Question: in TS/JS, is this passed by reference or by value? I.e., can we be sure that the selector is the same for all three objects afterwards?

portal/server/sentry.server.config.ts
```diff
@@ -7,7 +7,7 @@
 
 import * as Sentry from "@sentry/nextjs";
 
-if (process.env.ENABLE_SENTRY === "production") {
+if (process.env.ENABLE_SENTRY === "yes") {
```
giac-mysten, line 10: Question: what is behind this change?

portal/server/app/route.ts
giac-mysten: Question: do we still need this as a global singleton now?

portal/worker/src/walrus-sites-sw.ts
```diff
@@ -5,13 +5,28 @@ import { getDomain, getSubdomainAndPath } from "@lib/domain_parsing";
...
 // `WorkerGlobalScope & typeof globalThis` https://github.com/microsoft/TypeScript/issues/14877
 declare var self: ServiceWorkerGlobalScope;
 declare var clients: Clients;
 
+const rpcUrlList = process.env.RPC_URL_LIST;
+if (!rpcUrlList) {
+    throw new Error("Missing RPC_URL_LIST environment variable");
+}
+const rpcSelector = new RPCSelector(rpcUrlList.split(','));
+export const pageFetcher = new PageFetcher(
```
giac-mysten, line 24: Question: same here, do we need the singleton?

portal/common/lib/routing.ts
```diff
@@ -5,115 +5,123 @@ import { SuiObjectResponse } from "@mysten/sui/client";
...
+     * @returns The routes object.
+     */
+    private async fetchRoutesObject(objectId: string): Promise<SuiObjectResponse> {
+        return await this.rpcSelector.getObject({
+            id: objectId,
+            options: { showBcs: true },
         });
     }
-    throw new Error("Routes object data could not be fetched.");
-}
```
giac-mysten, line 50: Question: since we know the name of the dynamic field, why do we do two full node queries here, instead of computing the address of the dynamic field directly?

portal/common/lib/resource.ts
```diff
@@ -10,147 +10,151 @@ import { fromBase64 } from "@mysten/bcs";
...
-        );
-        return df.value;
+    /**
+    * Checks for loop detection and too many redirects.
+    * @param objectId
+    * @param seenResources
+    * @param depth
+    * @returns
+    */
+    private checkForErrors(
```
giac-mysten, line 138: Nit: the nome is not very descriptive. `checkRedirectLimits`?

giac-mysten (approved):
Awesome! this is definitely in the right direction! Thank you @Tzal3x . 
A few questions and suggestions on ways to improve further.

portal/common/lib/page_fetching.ts
Tzal3x: Nice idea, I will rename them. Better late than never!

portal/server/sentry.server.config.ts
Tzal3x: Ignore this, the correct on now after the refactoring is `ConfigurationLoader.config.enableSentry = true | false`.

portal/server/app/route.ts
Tzal3x: It's no longer a singleton in its' strict sense, meaning that its' constructor is no longer private, and multiple `rpcSelectors` can be created inside the portal. 

More specifically on these I am instantiating the `rpcSelector` once and pass it by reference to the other classes' constructors.

portal/worker/src/walrus-sites-sw.ts
Tzal3x: Let's track this conversation inside the thread above to avoid duplicate comments: https://github.com/MystenLabs/walrus-sites/pull/326#discussion_r1889032959

portal/common/lib/routing.ts
Tzal3x: Good catch! I am not sure why we missed this one in the past. I will look into it and replace it with `deriveDynamicField`. 🤔

portal/common/lib/page_fetching.ts
Tzal3x: Done, renamed `PageFetcher` to `UrlFetcher`, but I will keep the `ResouceFetcher` as is, because in the smart contract we have `site::Resource` and it's easier to map those two 1-1.

portal/common/lib/routing.ts
Tzal3x: I will add this in a later PR: https://github.com/MystenLabs/walrus-sites/pull/332

I did not made the commit to this branch because I was not able to test that the dynamic fields work due to the full node bug.

portal/common/lib/page_fetching.ts
giac-mysten: ah yes, that is whant i meant with "ResourceFetcher: as it is now"!

portal/server/app/route.ts
giac-mysten: ok, very interesting thanks!

I'm not familiar with these global `const`s, and i'm not sure if/how they could be problematic. but looks good for now i guess.

PR #327, opened 2024-12-13
feat(server-portal): create ConfigurationLoader

Tzal3x:
A utility class that makes safer to load the environment
variables of the project.

By using this class, we can ensure that the environment
variables are loaded correctly and their data types are
correct.


portal/server/configuration-loader.ts
giac-mysten, line 1: Nit: the name should be `configuration_loader.ts`

portal/server/configuration-loader.ts
```diff
@@ -0,0 +1,144 @@
...
+    portalDomainNameLength?: number;
+    premiumRpcUrlList: string[];
+    rpcUrlList: string[];
+    enableSentry: Boolean;
+    sentryAuthToken?: string;
+    sentryDsn?: string;
+}
+
+/**
+* A utility class that makes safer to load the environment
```
giac-mysten, line 29: Nit:
```suggestion
* A utility class that makes it safe to load the environment
```
and also line length

portal/server/configuration-loader.ts
```diff
@@ -0,0 +1,144 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+type StringBoolean = 'true' | 'false';
+function isStringBoolean(value: string): value is StringBoolean {
+    return value === 'true' || value === 'false';
+}
+function toBoolean(value: string): Boolean {
+    return value === 'true';
+}
```
giac-mysten, line 10: Question: I guess there is no way to store a private yaml file or similar instead of env variables?

giac-mysten (approved):
Awesome stuff :)

portal/server/configuration-loader.ts
Tzal3x: Yea there is no such built-in functionality in Vercel. Other custom options (such as using vercel's blob store to save yaml files, and then load them on the portal) seemed to overcomplicate the logic.

PR #333, opened 2024-12-18
feat: SEINT-275 Add http headers default cache-control to public, max-age=3600

VLegakis:


VLegakis:
After discussion with @Tzal3x we decided to leave this default cache-control policy for the future. Lets leave this PR open-pending and we decide depending on the type of sites being published.

Tzal3x:
@VLegakis, for now we decided not to include defaults, because we would like to align with current developer common practices. 

Take as an example vercel deployments where [caching headers are not added to the responses by default](https://vercel.com/docs/edge-network/headers/cache-control-headers#default-cache-control-value).

Closing this PR as it's no longer relevant. Thank you regardless! 🙇‍♂️

PR #334, opened 2024-12-18
feat(server-portal): unregister past service workers with script injection

Tzal3x:


Tzal3x:
This approach is wrong and it doesn't work. Closing the PR.

PR #336, opened 2024-12-22
fix: align response formats with Walrus

mlegner:
I modified some response formats in Walrus a bit carelessly in https://github.com/MystenLabs/walrus/pull/1324 and included this in the 1.7.3 release. Unfortunately this broke the site-builder... 😬 

Would be nice to merge this into `main` and `testnet`.


Tzal3x (approved):
Thank you for the quick fix @mlegner !

PR #337, opened 2024-12-24
feat(common-lib): derive routes df instead of fetch

Tzal3x:
This change will save us from doing one additional request each time we need to fetch the `Routes` dynamic field. 

It derives the dynamic field object ID of `Routes` instead of fetching it.

### DIFF
--- a/portal/common/lib/routing.ts
+++ b/portal/common/lib/routing.ts
@@ -7,6 +7,8 @@ import { DynamicFieldStruct, RoutesStruct } from "./bcs_data_parsing";
 import { bcs, fromBase64 } from "@mysten/bcs";
 import logger from "./logger";
 import { RPCSelector } from "./rpc_selector";
+import { deriveDynamicFieldID } from "@mysten/sui/utils";
+import { ROUTES_MOVE_TYPE } from "./constants";
 
 /**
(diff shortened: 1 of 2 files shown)


portal/common/lib/routing.ts
```diff
@@ -23,67 +24,47 @@ export class WalrusSitesRouter {
...
-     * @param client - The SuiClient instance.
-     * @param objectId - The ID of the dynamic field object.
+     * @param siteObjectId - The site object ID.
      * @returns The routes object.
      */
-    private async fetchRoutesObject(objectId: string): Promise<SuiObjectResponse> {
-        return await this.rpcSelector.getObject({
-            id: objectId,
+    private async fetchRoutesDynamicFieldObject(siteObjectId: string): Promise<SuiObjectResponse> {
+        const routesMoveType = "vector<u8>";
```
giac-mysten, line 57: Question: could/should this be a global const?

giac-mysten (approved):
Awesome! Thank you @Tzal3x !

portal/common/lib/routing.ts
Tzal3x: Since it's not used anywhere else, I didn't see any benefit making it public. We can make it public as soon as the need surfaces.

PR #340, opened 2025-01-04
fix: Correct range validation logic in isRangeValid function

Himess:
In the `isRangeValid` function, the previous check for `range.start !== null && range.end !== null` was incorrect. It has been updated to properly check if both `start` and `end` are `null` using `range.start == null && range.end == null`.

This ensures that a valid range is correctly determined when both `start` and `end` are `null`, and fixes the validation logic to work as intended.

- Fixed the logic to validate range properly.
- Added additional checks to ensure valid range values.
- Improved range validation to handle `null` and negative values correctly.


Tzal3x (approved):
Nice catch, thank you!

PR #343, opened 2025-01-07
fix(server-portal): return when Routes df is not found

Tzal3x:
There was an issue in the cases when a `Routes` dynamic field was not found, an exception would be thrown instead of showing a warning and returning `undefined` a.k.a. an implication that the df was _not found_.


giac-mysten (approved):
Thanks for the fix!

PR #344, opened 2025-01-08
feat: create SENTRY_TRACES_SAMPLE_RATE env variable

Tzal3x:
Defining the `SENTRY_TRACES_SAMPLE_RATE`
can give us more flexibility to increase and
decrease this number based on the traffic we get
without the need of creating PRs and changing
the code directly.


giac-mysten (approved):
🙌

PR #345, opened 2025-01-09
chore(site-builder): general reliability improvements

giac-mysten:
Add multiple reliability improvements to shield the site builder from network errors and full node rate limiting. 
Specifically:
- All calls to the sui client are now wrapped and are retried with backoff
- OS network failures (os error 54) are retried when fetching the dynamic fields

In a follow up, would be nice to start using `multi_get_object`.

### DIFF
--- a/site-builder/src/backoff.rs
+++ b/site-builder/src/backoff.rs
@@ -1,0 +1,238 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{num::Saturating, time::Duration};
+
+use rand::{rngs::StdRng, Rng, SeedableRng};
+use serde::{Deserialize, Serialize};
+
+/// Wrapper for the configuration for the exponential backoff strategy.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExponentialBackoffConfig {
+    /// The minimum backoff duration.
+    pub min_backoff: Duration,
+    /// The maximum backoff duration.
+    pub max_backoff: Duration,
+    /// The maximum number of retries.
+    ///
+    /// If `None`, the backoff strategy will keep retrying indefinitely.
+    pub max_retries: Option<u32>,
+}
+
+impl ExponentialBackoffConfig {
+    /// Creates a new configuration with the given parameters.
+    #[allow(dead_code)]
+    pub fn new(min_backoff: Duration, max_backoff: Duration, max_retries: Option<u32>) -> Self {
+        ExponentialBackoffConfig {
+            min_backoff,
+            max_backoff,
+            max_retries,
+        }
+    }
+
+    /// Gets a new [`ExponentialBackoff`] strategy with the given seed from the configuration.
+    pub fn get_strategy(&self, seed: u64) -> ExponentialBackoff<StdRng> {
+        ExponentialBackoff::new_with_seed(
+            self.min_backoff,
+            self.max_backoff,
+            self.max_retries,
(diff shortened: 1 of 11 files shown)


site-builder/src/retry_client.rs
```diff
@@ -0,0 +1,362 @@
...
+    site::contracts::{
+        get_sui_object_from_object_response,
+        AssociatedContractStruct,
+        TypeOriginMap,
+    },
+    types::SuiDynamicField,
+};
+
+/// The list of HTTP status codes that are retriable.
+const RETRIABLE_RPC_ERRORS: &[&str] = &["429", "500", "502"];
```
Tzal3x, line 51: What about the following? 
- 408: Request Timeout (bad request): The RPC timed out waiting for the request; might occur with large payloads.
- 503: Service Unavailable: The RPC is currently unable to handle the request.
- 504: Gateway Timeout: The server did not receive a timely response from an upstream server.

Also, 429 would need a bigger delay compared with others. Do we already do that?

site-builder/src/retry_client.rs
```diff
@@ -0,0 +1,362 @@
...
+            {
+                return true;
+            }
+        }
+        false
+    }
+}
+
+/// Retries the given function while it returns retriable errors.[
+async fn retry_rpc_errors<S, F, T, E, Fut>(mut strategy: S, mut func: F) -> Result<T, E>
```
Tzal3x, line 83: Noobie rust question: Is it a good pattern using that many generics in a function definition?

Tzal3x (approved):
Those are a lot changes! 🔥 Thanks @giac-mysten, I left a few minor questions.

site-builder/src/retry_client.rs
giac-mysten: interesting, adding these may make sense in a follow up yes. My reasoning was to add  only the error codes we actually witnessed in the wild for now, but these could be added and exposed through proper logging.

>Also, 429 would need a bigger delay compared with others. Do we already do that?

No, though i've seen that in practice it's not a problem (the request may fail a couple of times before the exponential backoff is large enough). This may also be a good followup.

site-builder/src/backoff.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+impl ExponentialBackoff<StdRng> {
+    /// Maximum number of milliseconds to randomly add to the delay time.
+    const MAX_RAND_OFFSET_MS: u64 = 1000;
+
+    /// Creates a new iterator with the provided minimum and maximum bounds,
+    /// and seeded with the provided value.
+    ///
+    /// If `max_retries` is `None`, this backoff strategy will keep retrying indefinitely.
```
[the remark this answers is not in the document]
giac-mysten: This is a generic backoff strategy, for which i think it makes sense to support indefinite retries (we use this in core for some tasks). Here, it's not a problem, since the end user can never specify this value to be `None`.

site-builder/src/retry_client.rs
giac-mysten: I created SEINT-518 to track this, closing for now.

site-builder/src/retry_client.rs
giac-mysten: Not sure if it's morally good, but it is a pattern i've encountered multiple times when dealing with futures and retries. The problem is that you need to express all the types of the returns (`Result<T, E>`), as well as the types of the function that generates the futures (`F -> Fut`) that need to be retried.

I would say it's hard to avoid in this case.

PR #346, opened 2025-01-09
feat(site-builder): add single resource to existing site

giac-mysten:
Introduces a new command, `update-resource`, with which a single resource can be added without verifying the state of the site object.


Tzal3x (approved):
Looks good to me! Thank you @giac-mysten! 🙏 💯

PR #359, opened 2025-01-17
fix(portal): service worker uninstall

giac-mysten:


portal/server/public/walrus-sites-sw.js
```diff
@@ -1,15 +1,21 @@
...
+        .then(() => self.clients.matchAll())
+        .then((clients) => {
+            clients.forEach((client) => {
+                if (client.url && "navigate" in client) {
+                    client.navigate(client.url);
+                }
+            });
+        });
+});
+Footer;
```
Tzal3x, line 21:
```suggestion
```

I think the `Footer` here should be removed.

giac-mysten: whooops that was a paste issue

PR #360, opened 2025-01-17
fix: log resolved suins when resp not null WIP

Tzal3x:
So far isValidResponse would check the responses
for completely different calls.

With this commit, I break it up into 3 distinct
functions that each check the adequate responses.

TODO
- Refactor
- Fix validity check for SuiNS response

### DIFF
--- a/portal/common/lib/rpc_selector.ts
+++ b/portal/common/lib/rpc_selector.ts
@@ -88,11 +88,13 @@ export class RPCSelector implements RPCSelectorInterface {
                 nullCurrentRPCClientUrl: this.selectedClient.getURL().toString()})
         }
 
-        if (this.isValidResponse(result)) {
-            return result;
-        } else {
-            throw new Error("Invalid response from selected client");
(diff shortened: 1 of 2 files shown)


portal/common/lib/rpc_selector.ts
```diff
@@ -129,45 +116,58 @@ export class RPCSelector implements RPCSelectorInterface {
             logger.info({ message: "RPC selected", rpcClientSelected: this.selectedClient.getURL() })
 
             return result;
-        } catch {
+        } catch (error) {
             throw new Error(`Failed to contact fallback RPC clients.`);
```
giac-mysten, line 120: Nit: do you use the error anywhere?

giac-mysten (commented):
LGTM! why the draft?

portal/common/lib/rpc_selector.ts
Tzal3x: Fixed 😁 !

giac-mysten (approved):

PR #362, opened 2025-01-20
feat(common): support for new SuiNS walrus site id resolution

Tzal3x:
### Summary

This PR integrates the new functionality of SuiNS name records supporting a `walrusSiteId` to enable sites to connect with both a wallet and a walrus site. For backwards compatibility, we fall back to `nameRecord.targetAddress`.

Note: We don't need to use a `rpcSelector.validateSuinsResponse` because if the full node fails to return something valid, it throws an error that we catch on an outer try/catch block. 

### Changelog
- install `@mysten/suins`
- upgrade `@mysten/sui` to `1.18.1` to be compatible with suins.
- add `getNameRecord` call to `WrappedSuiClient`.
- create a new `NameRecord` type in `common/types` since the suins lib does not export it.
- update `SuiNSResolver` to use `rpcSelector.getNameRecord`.
- update unit tests accordingly.

### DIFF
--- a/portal/common/lib/rpc_selector.ts
+++ b/portal/common/lib/rpc_selector.ts
@@ -88,14 +88,7 @@ export class RPCSelector implements RPCSelectorInterface {
                 nullCurrentRPCClientUrl: this.selectedClient.getURL().toString()})
         }
 
-        const isValid = methodName === "call" && args[0] === "suix_resolveNameServiceAddress" && this.isValidSuiNSResponse(result)
-        const isValidGetObject = methodName === "getObject" && this.isValidGetObjectResponse(result);
-        const isValidMultiGetObject = methodName === "multiGetObjects" && this.isValidMultiGetObjectResponse(result);
-        const isValidDynamicFieldObject = methodName === "getDynamicFieldObject" && this.isValidGetObjectResponse(result);
-        if (isValid || isValidGetObject || isValidMultiGetObject || isValidDynamicFieldObject) {
-            return result
-        }
-        throw new Error("Invalid response from selected client");
+        return result
     }
 
     // Fallback to querying all clients using Promise.any.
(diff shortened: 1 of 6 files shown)


portal/common/lib/rpc_selector.ts
```diff
@@ -8,32 +8,44 @@ import {
...
 class WrappedSuiClient extends SuiClient {
     private url: string;
+    private suinsClient: SuinsClient;
 
     constructor(url: string) {
         super({ url });
         this.url = url;
+        this.suinsClient = new SuinsClient({
+            client: this as SuiClient,
+            network: 'testnet' // TODO: get network from config
```
giac-mysten, line 32: Question: Will this be done in this pr? otherwise, open an issue and mention it here.

portal/common/lib/suins.ts
```diff
@@ -13,12 +14,15 @@ export class SuiNSResolver {
...
     ): Promise<string | null> {
-        const suiObjectId: string = await this.rpcSelector.call<string>("call", ["suix_resolveNameServiceAddress", [
-            subdomain + ".sui",
-        ]]);
-        if (suiObjectId) {
-            logger.info({ message: "resolved suins name", resolvedSuiNSName: subdomain, suiObjectId: suiObjectId });
-            return suiObjectId;
+        const nameRecord: NameRecord | null = await this.rpcSelector.getNameRecord(`${subdomain}.sui`);
+        if (nameRecord) {
+            const resolvedSuiNSName = nameRecord.walrusSiteId ?? nameRecord.targetAddress;
```
giac-mysten, line 19: Question: Could it be than neither of them is defined?

portal/common/lib/rpc_selector.ts
```diff
@@ -8,32 +8,44 @@ import {
...
+        });
     }
 
     public getURL(): string {
         return this.url;
     }
+
+    // Extends the SuiClient class to add a method to get a SuiNS record.
+    // Useful for treating the SuiClient as a SuiNS client during the invokeWithFailover method.
+    public async getNameRecord(name: string): Promise<NameRecord | null> {
```
giac-mysten, line 42: question: just to check i understand, this is just to have all the functionality to call to sui in the same class, correct?

giac-mysten (approved):
Thank you @Tzal3x , this looks good!

portal/common/lib/suins.ts
Tzal3x: No because if neither of them is defined, there will be no name record. 
This is also depicted to the NameRecord declaration, where targetAddress must be always defined:
```ts
export type NameRecord = {
	name: string;
	nftId: string;
	targetAddress: string;
	expirationTimestampMs: number;
	data: Record<string, string>;
	avatar?: string;
	contentHash?: string;
	walrusSiteId?: string;
};
```

portal/common/lib/rpc_selector.ts
Tzal3x: https://github.com/MystenLabs/walrus-sites/issues/364

portal/common/lib/rpc_selector.ts
Tzal3x: Correct! 💯

PR #373, opened 2025-01-29
refactor: move source files to a src/ directory

Tzal3x:
tl;dr: just move a bunch of files to `src/`

Separates the configuration/setup files and
middleware code from the source files.


giac-mysten (approved):
Cool 🙏 can we also separate out the tests?

PR #374, opened 2025-01-29
feat: sew 21 migrate the allowlist from vercel edge config to redis

Tzal3x:
## Changelog
- Create an AllowlistChecker, similar to the BlocklistChecker. 
- Create a RedisClientFacade class that contains all the shareable Redis functions.


portal/server/src/allowlist_checker.ts
giac-mysten, line 1: Question: Isn't this the same code as the blocklist? can we not bring them together more closely? e.g. by differentiating only the variants map, and passing the map to a common builder?

giac-mysten (approved):
LGTM, generally there is a bit of duplicate code that could be removed? Otherwise looks good!

portal/server/src/allowlist_checker.ts
Tzal3x: I did this initially, but in the end I preferred reverting to the duplicated code because a common `CheckerBuilder` complicated the code too much. 

Code duplication for a few instances is [not that bad](https://stackoverflow.com/a/1765927). 

The duplicates felt easier to work with, and it's better that the code between the blocklist and allowlist is completely independent, since requirements most likely would diverge them even more. 

I would prefer optimizing for a common builder later on, if that is necessary.

PR #376, opened 2025-01-30
refactor: replace custom config loader with Zod

Tzal3x:
This is a more clean approach for environment variable parsing. 

We can use [Zod](https://zod.dev/) to validate if all of the environment variables are set, and to automatically get types for them. Plus define "refinements", i.e. complex validity checks for each environment variable.

For example, to check that the `RPC_URL_LIST="https://sui-testnet.public.blastapi.io,https://testnet.suiet.app"` is parsed correctly:

```ts
rpcUrlList: z.preprocess(
// Check that the value is a string, trim it and split it in order to get an array.
(val) => typeof val === 'string' ? val.trim().split(',') : val,
// Verify that the result is indeed an array, each is element is a string and a valid URL. 
z.array(z.string().url())
)
```


Tzal3x:
Something is wrong with my formatting 🤦‍♂️ ...

chariskms (approved):
LGTM. Zod seems like a cleaner solution.

PR #377, opened 2025-01-30
feat: sew 25 replace node runtime with bun

Tzal3x:
The core feature of Bun is that it uses the JavascriptCore instead of V8.
It's a drop-in replacement for node.js. 

Benefits
- [at least 4x faster startup time compared to Node.js](https://bun.sh/docs#design-goals)
- faster build times
- huge boost to our local development experience

## Changelog
- Replace `pnpm` with `bun` wherever it applied.
- Update github actions
- Remove the Codspeed integration. There is no benefit having it around, we were only using it for file decompression checks. 
- Update `package.json` files. A lot of scripts defined there are not used, so why keep them around? 
- Move static files to the service worker: https://github.com/MystenLabs/walrus-sites/pull/379


portal/server/next-env.d.ts
```diff
@@ -1,6 +1,3 @@
-// Copyright (c) Mysten Labs, Inc.
-// SPDX-License-Identifier: Apache-2.0
-
```
giac-mysten, line 3: Nit: the header should be here. Did you run the pre-commit?

portal/server/package.json
```diff
@@ -14,9 +14,11 @@
         "@vercel/edge-config": "^1.4.0",
         "common": "workspace:common",
         "next": "14.2.21",
-        "redis": "^4.7.0"
+        "redis": "^4.7.0",
+        "zod": "^3.24.1"
```
giac-mysten, line 18: Question: shouldn't this be in another PR?

portal/common/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+  	"allowSyntheticDefaultImports": true,
```
giac-mysten, line 3: Question: why do we need this?

portal/worker/pnpm-lock.yaml
giac-mysten, line 1: Question: why is this pnpm lock still around? remove?

portal/worker/README.md
```diff
@@ -15,8 +15,8 @@ involves loading objects from chain, decoding them, and serving them as response
 A local portal can be run from this directory using `pnpm`:
 
 ```shell
-pnpm install
-pnpm serve
+bun install
+bun serve:prod
```
giac-mysten, line 19: Q: why prod?

portal/vercel.json
```diff
@@ -1,6 +1,7 @@
 {
     "framework": null,
-    "installCommand": "pnpm install",
-    "buildCommand": "pnpm build:worker",
+    "installCommand": "bun install",
+    "buildCommand": "bun run build:worker",
+    "devCommand": "bun run worker",
```
giac-mysten, line 5: Q: do we want to switch to server?

giac-mysten (commented):
Made a first pass of comments 🙏

portal/vercel.json
Tzal3x: This `vercel.json` is only used in `walrus-sites-sw-testnet` (service worker). 

The corresponding `bun run start` is defined in the `portal/server/vercel.json`.

I guess we can improve this in a following PR. I didn't change the file structure of `vercel.json` files in order to avoid making changes to the existing vercel configs in the UI.

portal/worker/README.md
Tzal3x: Prod maps to `"serve:prod": "webpack serve --config webpack.config.prod.js"`, which is the correct one.

Just serve, would serve the `dev` service worker which is suboptimal.

portal/worker/pnpm-lock.yaml
Tzal3x: That was a merge conflict. Should be fine now.

portal/common/tsconfig.json
Tzal3x: The language server was complaining that in order to import html files we needed to define this.

portal/server/package.json
Tzal3x: It was but we pointed to this branch instead of main 😓 . 

https://github.com/MystenLabs/walrus-sites/pull/376

portal/server/next-env.d.ts
Tzal3x: Done! I had added it to `.gitignore` temporarily and it wouldn't pass the pre-commit. 

Initially I added `next-env.d.ts` to the `.gitignore` because it would be re-written without the contents of it changing, but it would require the pre-commit to run always)

Should be fine now. 👍

portal/worker/README.md
Tzal3x: Yes that could work too, but I preferred this to be explicit, since the past mistake was due to this implicitness.

giac-mysten (approved):
LGTM, let's try this out!

PR #378, opened 2025-01-30
chore: fix typo in concrete.css

teenager-ETH:


Tzal3x:
Thank you for the contribution @teenager-ETH ! However, we are making a lot of changes in the project's structure and this commit will create conflicts. Plus, `concrete.css` will be removed very soon so I would prefer to proceed without merging this. 

Thank you nevertheless, and feel free to submit another fix if you find anything! 💯

PR #385, opened 2025-02-03
feat: retry fetch aggregator on exception

chariskms:
Added a retry mechanism for fetching resources from the aggregator.

Changes:

- Introduced fetchWithRetry method to retry failed fetch requests up to 3 times with a delay.
- Updated UrlFetcher to use fetchWithRetry instead of direct fetch calls.

### DIFF
--- a/portal/common/lib/url_fetcher.ts
+++ b/portal/common/lib/url_fetcher.ts
@@ -159,7 +159,7 @@ export class UrlFetcher {
         // We have a resource, get the range header.
         logger.info({ message: "Add the range headers of the resource", range: JSON.stringify(result.range)});
         let range_header = optionalRangeToRequestHeaders(result.range);
-        const contents = await fetch(
+        const contents = await this.fetchWithRetry(
             aggregatorEndpoint(result.blob_id, this.aggregatorUrl), { headers: range_header }
         );
         if (!contents.ok) {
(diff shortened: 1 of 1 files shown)


portal/common/lib/url_fetcher.ts
```diff
@@ -192,4 +192,42 @@ export class UrlFetcher {
...
         });
     }
+
+    /**
+     * Attempts to fetch a resource from the given input URL or Request object, with retry logic.
+     *
+     * Retries the fetch operation up to a specified number of attempts in case of failure,
+     * with a delay between each retry. Logs the status and error messages during retries.
+     *
+     * @param input - The URL, URL object, or Request object representing the resource to fetch.
```
Tzal3x, line 202:
```suggestion
     * @param input - The URL string, URL object, or Request object representing the resource to fetch.
```

portal/common/lib/url_fetcher.ts
```diff
@@ -192,4 +192,42 @@ export class UrlFetcher {
...
+     * @param init - Optional fetch options such as headers, method, and body.
+     * @param retries - The maximum number of retry attempts (default is 3).
+     * @param delayMs - The delay in milliseconds between retry attempts (default is 1000ms).
+     * @returns A promise that resolves with the successful `Response` object or rejects with the last error.
+     */
+    private async fetchWithRetry(
+        input: string | URL | globalThis.Request,
```
Tzal3x, line 209: Question: why do we have to define this like `globalThis.Request` instead of just `Request`?

portal/common/lib/url_fetcher.ts
```diff
@@ -192,4 +192,42 @@ export class UrlFetcher {
...
+                lastError = error;
+            }
+
+            // Wait before retrying
+            if (attempt < retries) {
+                await new Promise(resolve => setTimeout(resolve, delayMs));
+            }
+        }
+        // All retry attempts failed; throw the last encountered error.
+        throw lastError instanceof Error ? lastError : new Error('Unknown error occurred in fetchWithRetry');
```
Tzal3x, line 231: If we expect `lastError` to be an instanceof Error, then why have we defined it as `let lastError: unknown`?

Tzal3x (changes requested):
Nice, left a few. 💬

portal/common/lib/url_fetcher.ts
chariskms: If an invalid value for retries is set, it might never enter the loop, so we should ensure it logs an error in that case.
Maybe adding an if check and falling back to a generic fetch would be a better approach? This way, we handle incorrect retries more gracefully.

chariskms: It's generally good practice to normalize any thrown exceptions to ensure they are instances of [[Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error). However, since the Fetch API only throws errors for network-related issues or when a request is aborted (as detailed in the [[Using Fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch)](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) documentation), we can be confident that only Error objects will be thrown in this case.

portal/common/lib/url_fetcher.ts
chariskms: It is the definition of fetch.

`function fetch(
        input: string | URL | globalThis.Request,
        init?: RequestInit,
    ): Promise<Response>;`

PR #386, opened 2025-02-03
feat: sew 26 amplitude integration along with web analytics

Tzal3x:
Integrate amplitude event tracking with the aim to gradually migrate from vercel web analytics.

Currently the use of Amplitude doesn't disable web analytics in order for us to make a smooth transition.

### DIFF
--- a/portal/server/src/amplitude.ts
+++ b/portal/server/src/amplitude.ts
@@ -1,0 +1,53 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import * as amplitude from "@amplitude/analytics-node";
+import { generateHash } from "./utils";
+import { config } from "./configuration_loader";
+import logger from "@lib/logger";
+
+if (config.amplitudeApiKey) {
+	amplitude.init(process.env.AMPLITUDE_API_KEY!,{
+		// Events queued in memory will flush when number of events exceed upload threshold
+		// Default value is 30
+		flushQueueSize: 50,
+		// Events queue will flush every certain milliseconds based on setting
+		// Default value is 10000 milliseconds
+		flushIntervalMillis: 5000, // TODO increase this to 20000 for production use.
+	});
+}
+
+/**
+* Sends a page view event to Amplitude.
+* @param req - The incoming request to the portal.
+*/
+export async function sendToAmplitude(req: Request): Promise<void> {
+	if (!config.amplitudeApiKey) {
+		logger.warn({ message: "Amplitude API key not found. Skipping tracking." });
+		return;
+	}
+	try {
+		amplitude.track({
+			device_id: generateDeviceId(req.headers.get("user-agent")),
+	    	event_type: "page_view",
+	      	event_properties: {
+	        	url: req.url,
+	          	referrer: req.headers.get("referer") ?? "direct",
+		        user_agent: req.headers.get("user-agent") ?? undefined,
+			},
+      		ip: req.headers.get("x-forwarded-for") ?? undefined, // TODO: Parse using a lib
+  	    })
+	} catch (e) {
+		console.warn("Amplitude could not track event: ", e);
+	}
+}
+
+/**
+* Generates a device ID based on the user agent string.
+* @param userAgent - device & browser details. Default is a random string.
+* @returns A hashed device ID.
+*/
+function generateDeviceId(userAgent: string | null): string {
+	const defaultDeviceId = "1234567890";
+	return generateHash(userAgent || defaultDeviceId);
+}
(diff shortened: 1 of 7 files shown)


portal/server/src/amplitude.ts
```diff
@@ -0,0 +1,51 @@
...
+import logger from "@lib/logger";
+import { NextRequest } from "next/server";
+
+if (config.amplitudeApiKey) {
+	amplitude.init(process.env.AMPLITUDE_API_KEY!,{
+		// Events queued in memory will flush when number of events exceed upload threshold
+		// Default value is 30
+		flushQueueSize: 50,
+		// Events queue will flush every certain milliseconds based on setting
+		// Default value is 10000 milliseconds
```
giac-mysten, line 16:
```suggestion
		// Events queue will flush every certain milliseconds based on setting.
		// Default value is 10000 milliseconds.
```

portal/server/src/amplitude.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as amplitude from "@amplitude/analytics-node";
+import { generateHash, isHtmlPage } from "./utils";
+import { config } from "./configuration_loader";
+import logger from "@lib/logger";
+import { NextRequest } from "next/server";
+
+if (config.amplitudeApiKey) {
+	amplitude.init(process.env.AMPLITUDE_API_KEY!,{
+		// Events queued in memory will flush when number of events exceed upload threshold
+		// Default value is 30
```
giac-mysten, line 13: Nit
```suggestion
		// Events queued in memory will flush when number of events exceed upload threshold.
		// Default value is 30.
```

portal/server/src/amplitude.ts
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+/**
+* Generates a device ID based on the user agent string.
+* @param userAgent - device & browser details. Default is a random string.
+* @returns A hashed device ID.
+*/
+function generateDeviceId(userAgent: string | null): string {
+	const defaultDeviceId = "1234567890";
```
giac-mysten, line 49: Remark: this is not a random string?

portal/server/src/amplitude.ts
```diff
@@ -0,0 +1,51 @@
...
+			device_id: generateDeviceId(request.headers.get("user-agent")),
+	    	event_type: "page_view",
+  	    })
+	} catch (e) {
+		console.warn("Amplitude could not track event: ", e);
+	}
+}
+
+/**
+* Generates a device ID based on the user agent string.
```
giac-mysten, line 44: Question: how specific must this be? does the user agent have enough entropy to distinguish individual clients? I'm guessing no?

portal/server/src/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    // This is to avoid tracking requests for static assets like images, css, etc.
+    // Cuts down costs since we are tracking less events.
+    const parsedUrl = getSubdomainAndPath(
+        request.nextUrl,
+        Number(config.portalDomainNameLength)
+    );
+    if (!parsedUrl?.path) {
+    	throw new Error("No path found in parsed URL");
+    }
+    return parsedUrl?.path?.endsWith('.html')
```
giac-mysten, line 26: Important: this will not work. Html pages can have multiple extensions (`htm` and `html` are the most common). But even more, the url may not have an extension and still load html. Can we look at the content-type header instead?

giac-mysten (changes requested):
Thank you @Tzal3x , made a quick first pass

portal/server/src/amplitude.ts
Tzal3x: It shouldn't be completely random in order to track the same device again when it gets used again for a future request.

portal/server/src/amplitude.ts
Tzal3x: Each client is identified using a user id. Each user id has a few devices - defined by a device id. We can test it out.

portal/server/src/utils.ts
Tzal3x: Done, kept `parsedUrl?.path?.endsWith('.html')` as a fallback if content-type is not defined for some reason.

portal/server/src/amplitude.ts
giac-mysten: ok, then don't call it random in the docstring :)

giac-mysten (approved):
Nice!

portal/server/src/amplitude.ts
Tzal3x: I found this registry with [examples of user agents](https://deviceatlas.com/blog/list-of-user-agent-strings). I think they include enough entropy.

PR #399, opened 2025-02-11
chore: check that each redis url db number is defined correctly

Tzal3x:
Update the configuration loader zod parsing checks so that when redis urls are provided for the blocklist and allowlist, each database number should be defined correctly. This acts as a Ulysses pact, to prevent us from switching accidentally the db numbers.

i.e. we use the same redis instance for both databases. However, for each database a number should be defined.
The blocklist database should be database `0` and the allowlist the database `1`. 

e.g. 
- `BLOCKLIST_REDIS_URL = redis[s]://[[same-username][:same-password]@][same-host][:same-port][/0]`
- `ALLOWLIST_REDIS_URL = redis[s]://[[same-username][:same-password]@][same-host][:same-port][/1]`

### DIFF
--- a/portal/server/next-env.d.ts
+++ b/portal/server/next-env.d.ts
@@ -1,5 +1,6 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
+
 /// <reference types="next" />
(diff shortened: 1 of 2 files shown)


giac-mysten (approved):
thanks @Tzal3x !

PR #402, opened 2025-02-14
feat(common): update constants to support different networks

Tzal3x:
We need to be able to be able to deploy a portal that serves sites based on a different network without the need of changing the code directly. 

To do this:
- [x] Remove AGGREGATOR and SITE_PACKAGE from `constants.ts` and parse it dynamically wherever needed. 
- [x] Create a new env var - The aggregator url (e.g. `https://aggregator.walrus-testnet.walrus.space`).
- [x] Create a new env var - The SITE_PACKAGE - we can have different packages for different kinds of networks (mainnet/testnet).
- [x] Create a PR to sui-operations for the next deployments to work. https://github.com/MystenLabs/sui-operations/pull/4976
- [x] test, test, **TEST**!


portal/common/lib/constants.ts
```diff
@@ -1,8 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-export const NETWORK = "testnet";
-export const AGGREGATOR = "https://aggregator.walrus-testnet.walrus.space";
+// TODO(alex): add MAINNET_SITE_PACKAGE
 export const SITE_PACKAGE = "0xc5bebae319fc9d2a9dc858b7484cdbd6ef219decf4662dc81a11dc69bb7a5fa7";
```
giac-mysten, line 5: question: are we not storing the site package in an env var now?

giac-mysten (approved):
LGTM!

portal/common/lib/constants.ts
Tzal3x: Nice catch, this was not intentional. 👍

Tzal3x: Removed them in [27ffeed](https://github.com/MystenLabs/walrus-sites/pull/402/commits/27ffeed45db8e9e820060216cac0bfc5a2e00f07).

PR #404, opened 2025-02-19
fix: do not set portal length by default

pei-mysten:
this is causing the `splitUrl` to not work as expected for `walurs.site` and `blob.store`. super easy fix with quite a lengthy debug process...
although now we won't have `walrus-sites-testnet.mystelanbs.com` working as the `splitUrl` would see the subdomain as something like `stake.walrus-sites-testnet` instead of just `stake`. But still, it's way more important to make sure `walrus.site` and `blob.store` work


Tzal3x (approved):
Nice catch, thank you @pei-mysten !

PR #405, opened 2025-02-20
fix: amplitude user agent header

pei-mysten:
amplitude is kind of broken after being moved behind a LB
1. `request.nextUrl` is not being reliable behind a LB, it's showing something like `http://localhost:3000/something` in the log, we cannot infer subdomain from `localhost`
2. `ua = new uaparser.UAParser(request.headers.get("user-agent") ?? undefined);` will always fail as the `UAParser` due to constructor overload. By always passing a string to the constructor, we can avoid the confusion.
3. `request.headers.get("x-forwarded-for")` are something like `ip1, ip2` behind GCP LB, we should always take the first ip in the chain, otherwise Amplitude will be confused and show completely incorrect geo country.

### DIFF
--- a/portal/docker/server/Dockerfile
+++ b/portal/docker/server/Dockerfile
@@ -59,4 +59,4 @@ COPY --from=install /temp/prod/server/.next server/.next
 USER bun
 EXPOSE 3000/tcp
 ENV NODE_ENV=production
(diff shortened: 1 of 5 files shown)


Tzal3x (approved):
Awesome, thank you! 💯

PR #406, opened 2025-02-20
fix(site-builder): fix parsing of blob extensions

mlegner:
Recently, support for blob extensions was added to the `walrus` CLI, this PR makes the site-builder compatible with these new outputs. See [here](https://github.com/MystenLabs/walrus-docs/actions/runs/13430256831/job/37537274187) for an example of the incompatibility.

This should be cherry-picked to the `testnet` branch.


giac-mysten (approved):
🙏

PR #414, opened 2025-02-28
chore: fix dockerfile

pei-mysten:
1. `over/bun:1` will automatically use latest 1.x.x release, which is 1.2.4 now, but it has some compatibility issues with the current version of `zod` we're using, so fixing it to 1.2.1 as indicated in bun.lock
2. make base image cleaner, those env vars are needed only during building phase, so no need to be in the base image.
3. copy `public/` folder over so that we can serve `walrus-site-sw.js` for some 404 issues


Tzal3x (approved):
lgtm, thanks @pei-mysten 💯

PR #416, opened 2025-03-01
fix: make encoding type compatible with latest Walrus version

mlegner:
This needs to be merged in both `main` and `testnet`.


Tzal3x (approved):
Thank you @mlegner!

PR #419, opened 2025-03-04
feat(contract): sew 58 display object for walrus sites nft

Tzal3x:


move/walrus_site/sources/site.move
```diff
@@ -212,4 +243,44 @@ module walrus_site::site {
         } = site;
         id.delete();
     }
+
+    #[allow(lint(self_transfer))]
+    /// Define a Display for the Site objects.
+    public fun set_site_display(publisher: &Publisher, ctx: &mut TxContext) {
+        // Check if the `Publisher` has the authority over the Display.
+        // Checks if the type is from the same module, hence the
```
giac-mysten, line 251: Nit: unfinished docs

move/walrus_site/sources/site.move
```diff
@@ -212,4 +243,44 @@ module walrus_site::site {
         } = site;
         id.delete();
     }
+
+    #[allow(lint(self_transfer))]
+    /// Define a Display for the Site objects.
+    public fun set_site_display(publisher: &Publisher, ctx: &mut TxContext) {
```
giac-mysten, line 249: Question: what is the goal of having this as a public function? future upgrades? In general, new public functions can be added with upgrades.

move/walrus_site/sources/site.move
```diff
@@ -212,4 +243,44 @@ module walrus_site::site {
...
+        // Check if the `Publisher` has the authority over the Display.
+        // Checks if the type is from the same module, hence the
+        assert!(publisher.from_module<Site>(), ENotAuthorized);
+        // Checks if the type is from the same package.
+        assert!(publisher.from_package<Site>(), ENotAuthorized);
```
giac-mysten, line 254: Question: are these checks necessary? if this function becomes only something used in the initial setup, they probably are not.

move/walrus_site/sources/site.move
```diff
@@ -10,11 +10,18 @@ module walrus_site::site {
...
     /// The site published on Sui.
     public struct Site has key, store {
         id: UID,
         name: String,
+        link: Option<String>,
+        image_url: Option<String>,
+        description: Option<String>,
+        project_url: Option<String>,
+        creator: Option<String>,
+        metadata: Option<String>,
```
giac-mysten, line 24: Question: what is this field for?

move/walrus_site/sources/site.move
```diff
@@ -52,11 +59,35 @@ module walrus_site::site {
...
+        set_site_display(&publisher, ctx);
+        transfer::public_transfer(publisher, ctx.sender());
+    }
+
     /// Creates a new site.
-    public fun new_site(name: String, ctx: &mut TxContext): Site {
+    public fun new_site(
+        name: String,
+        link: Option<String>,
+        image_url: Option<String>,
```
giac-mysten, line 75: Suggestion: can we default the `image_url` to `https://www.walrus.xyz/walrus-site`?

move/walrus_site/sources/site.move
```diff
@@ -10,11 +10,18 @@ module walrus_site::site {
     const EResourceDoesNotExist: u64 = 0;
     const ERangeStartGreaterThanRangeEnd: u64 = 1;
     const EStartAndEndRangeAreNone: u64 = 2;
+    const ENotAuthorized: u64 = 3;
 
     /// The site published on Sui.
     public struct Site has key, store {
```
giac-mysten, line 16: Question: I guess we now need setters?

giac-mysten (commented):
Thanks! a few preliminary comments

move/walrus_site/sources/site.move
damirka: +1 to the question.

move/walrus_site/sources/site.move
damirka, line 249: I don't think this is the right way to do it. You'll end up creating Display objects for every update, and instead, one should have 1 display object with updates in this object.

move/walrus_site/sources/site.move
```diff
@@ -212,4 +243,44 @@ module walrus_site::site {
         } = site;
         id.delete();
     }
+
+    #[allow(lint(self_transfer))]
+    /// Define a Display for the Site objects.
+    public fun set_site_display(publisher: &Publisher, ctx: &mut TxContext) {
+        // Check if the `Publisher` has the authority over the Display.
+        // Checks if the type is from the same module, hence the
+        assert!(publisher.from_module<Site>(), ENotAuthorized);
```
damirka, line 252: Makes no sense, this check is already performed in `display::new`. Same for below.

move/walrus_site/sources/site.move
damirka, line 254: if `from_module` is true, `from_package` is automatically true, check is redundant.

move/walrus_site/sources/site.move
```diff
@@ -52,11 +59,35 @@ module walrus_site::site {
...
+        let publisher = package::claim(otw, ctx);
+        set_site_display(&publisher, ctx);
+        transfer::public_transfer(publisher, ctx.sender());
+    }
+
     /// Creates a new site.
-    public fun new_site(name: String, ctx: &mut TxContext): Site {
+    public fun new_site(
+        name: String,
+        link: Option<String>,
```
damirka, line 74: I think all of them should be moved into a single `Metadata` field which can be edited / set / updated in a single call or copied over. Too many arguments is not a good strategy for usability.

move/walrus_site/sources/site.move
```diff
@@ -212,4 +243,44 @@ module walrus_site::site {
...
+            b"{name}".to_string(),
+            b"{link}".to_string(),
+            b"{image_url}".to_string(),
+            b"{description}".to_string(),
+            b"{project_url}".to_string(),
+            b"{creator}".to_string(),
+            b"{metadata}".to_string(),
+        ];
+
+        let mut display = display::new_with_fields<Site>(
```
damirka, line 276: We want to guarantee there's only 1 display in the system to later add it to display registry.

move/walrus_site/sources/site.move
damirka: Add a separate module with a type `public struct Metadata { fields } has copy, drop, store` and accept it as an argument here.

move/walrus_site/sources/site.move
```diff
@@ -1,7 +1,7 @@
 /// The module exposes the functionality to create and update Walrus sites.
 module walrus_site::site {
```
damirka, line 2: Use module label.

move/walrus_site/sources/site.move
damirka: So you probably want to make this private and call it in the `init` instead.

move/walrus_site/sources/site.move
Tzal3x: The idea was to include a metadata field that acts as a placeholder for additional data that the user wants to include. For example, github commit of the site, tags that describe the project etc.

move/walrus_site/sources/site.move
Tzal3x: Changed it to private as @damirka suggested. ✅

move/walrus_site/sources/site.move
Tzal3x: `set_site_display` is now private and is only called inside the `init` function so this should be resolved now.

move/walrus_site/sources/site.move
Tzal3x: Done!

sites-config.yaml
```diff
@@ -1,6 +1,6 @@
 # module: site
 # portal: walrus.site
-package: 0xdf9033cac39b7a9b9f76fb6896c9fc5283ba730d6976a2b1d85ad1e6036c3272
+package: 0xe004477d29e97cc42fa00203227fad8d87eabf583b79618dddfde638e8efc11c
```
Tzal3x, line 3: This is only temporary so that reviewers can try out the code. 

The `package` here should be updated when we redeploy the contracts before mainnet.

move/walrus_site/sources/site.move
Tzal3x: Created an `update_metadata` function that does this.

move/walrus_site/sources/site.move
Tzal3x: Since we are using move prettier, when I add the `--write` flag, it goes back to the bracket syntax.

move/walrus_site/sources/events.move
```diff
@@ -0,0 +1,51 @@
+module walrus_site::events {
```
damirka, line 1: Nit: fmt to add module label.

move/walrus_site/sources/metadata.move
```diff
@@ -0,0 +1,49 @@
...
+        Metadata {
+            link,
+            image_url,
+            description,
+            project_url,
+            creator,
+        }
+    }
+
+    public fun link(metadata: &Metadata): Option<String> {
```
damirka, line 30: Do you want to add setters?

move/walrus_site/sources/site.move
damirka: Then add a configuration:

prettierrc.json
```
{
    "useModuleLabel": true,
    "autoGroupImports": "module"
}
```

move/walrus_site/sources/site.move
```diff
@@ -206,10 +250,58 @@ module walrus_site::site {
...
     public fun burn(site: Site) {
+        emit_site_burned(object::id(&site));
         let Site {
             id,
             ..,
         } = site;
         id.delete();
     }
+
+    #[allow(lint(self_transfer))]
```
damirka, line 261: Lint makes no sense given return value.

move/walrus_site/sources/site.move
```diff
@@ -206,10 +250,58 @@ module walrus_site::site {
...
+        d
+    }
+
+    #[test_only]
+    public fun init_for_testing(ctx: &mut TxContext) {
+        init(SITE {}, ctx);
+    }
+
+    #[test_only]
+    public fun get_site_name(site: &Site): String {
```
damirka, line 299: Why is it a test-only? Is there a public getter? Same for below.

move/walrus_site/tests/site_tests.move
```diff
@@ -146,4 +166,57 @@ module walrus_site::site_tests {
...
+            let mut site = scenario.take_from_sender<Site>();
+            // Update the site metadata.
+            let metadata = walrus_site::metadata::new_metadata(
+                option::some(b"https://<b36>.walrus.site".to_string()),
+                option::some(b"https://<b36>.walrus.site/image.png".to_string()),
+                option::some(b"I am just updating the site metadata.".to_string()),
+                option::none(),
+                option::none(),
+            );
+            walrus_site::site::update_metadata(&mut site, metadata);
```
damirka, line 207: nit:
```suggestion
            site.update_metadata(metadata);
```

move/walrus_site/sources/site.move
```diff
@@ -1,7 +1,11 @@
 /// The module exposes the functionality to create and update Walrus sites.
 module walrus_site::site {
```
damirka, line 2: Missing public getters: metadata, name

Tzal3x: If we don't use them anywhere currently, do we have to include them? Shouldn't we follow YAGNI?

Tzal3x: > If you don't provide getter for metadata - what's the point to provide metadata::* getters?

The metadata::* getters are used as an alternative to access the Metadata fields, because we can't deconstruct an object of a struct that is defined in a separate module.

We don't store the whole `Metadata` object anywhere; it's just a helper struct to make the `new_site` function to have less arguments as recommended here: https://github.com/MystenLabs/walrus-sites/pull/419#discussion_r1979430414. 

These metadata fields have to be parseable to the Display object, that's why we don't store the whole metadata object under the `Site` object:

```move
            b"{name}".to_string(),
            b"{link}".to_string(),
            b"{image_url}".to_string(),
            b"{description}".to_string(),
            b"{project_url}".to_string(),
            b"{creator}".to_string(),
```

But I can add the setters inside the `site::metadata` to make it easier for 3rd parties as you suggested. Will do the same for

move/walrus_site/sources/events.move
```diff
@@ -0,0 +1,51 @@
...
+use walrus_site::metadata::Metadata;
+
+public struct SiteCreatedEvent has copy, drop {
+    site_id: ID,
+    site_name: String,
+    site_metadata_link: Option<String>,
+    site_metadata_image_url: Option<String>,
+    site_metadata_description: Option<String>,
+    site_metadata_project_url: Option<String>,
+    site_metadata_creator: Option<String>,
```
giac-mysten, line 14: Suggestion: I'd remove allo of these parameters. They are probably not very useful, and can be inferred by the created objects if needed.

giac-mysten (approved):
Awesome, thank you @Tzal3x

.github/workflows/move-tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
       - name: Download Sui
         run: |
-          VERSION=$(curl -s https://api.github.com/repos/MystenLabs/sui/releases/latest | jq -r '.tag_name')
+          VERSION=testnet-v1.43.1
```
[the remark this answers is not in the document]
Tzal3x: It was necessary because there was a breaking change with the latest version (after v.144) and the builds would fail.

move/walrus_site/.prettierrc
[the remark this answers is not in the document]
Tzal3x: No this only applies to move!

move/walrus_site/sources/events.move
```diff
@@ -0,0 +1,51 @@
...
+
+public struct SiteBurnedEvent has copy, drop {
+    site_id: ID,
+}
+
+public struct SiteNameUpdateEvent has copy, drop {
+    site_id: ID,
+    old_name: String,
+    new_name: String,
+}
```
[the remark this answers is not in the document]
Tzal3x: Yes it should be redundant, I will remove it.

PR #424, opened 2025-03-06
feat(portal): sew 82 testnet portal until mainnet should not support b36 domains

Tzal3x:
- Create a B36_DOMAIN_RESOLUTION_SUPPORT environment variable.
- Integrate it to the UrlFetcher class.
- The landing page b36 will always resolve to an object Id regardless of B36_DOMAIN_RESOLUTION_SUPPORT.
- Add B36_DOMAIN_RESOLUTION_SUPPORT to the server/Dockerfile.
- ⚠️ The service worker will always support b36 domains. 
- Update the site-builder publish output string. (`Browse the site with ...`).

Related PR on sui-operations: https://github.com/MystenLabs/sui-operations/pull/5078

### DIFF
--- a/portal/common/lib/url_fetcher.ts
+++ b/portal/common/lib/url_fetcher.ts
@@ -23,7 +23,6 @@ import { WalrusSitesRouter } from "./routing";
 import { HttpStatusCodes } from "./http/http_status_codes";
 import logger from "./logger";
 import BlocklistChecker from "./blocklist_checker";
-import {Network} from "@mysten/suins";
 
 /**
 * Includes all the logic for fetching the URL contents of a walrus site.
@@ -34,6 +33,7 @@ export class UrlFetcher {
         private suinsResolver: SuiNSResolver,
         private wsRouter: WalrusSitesRouter,
         private aggregatorUrl: string,
+        private b36DomainResolutionSupport: boolean
     ){}
 
     /**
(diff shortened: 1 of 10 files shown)


portal/common/lib/url_fetcher.ts
```diff
@@ -121,37 +121,38 @@ export class UrlFetcher {
...
         parsedUrl: DomainDetails,
     ): Promise<string | Response> {
-        let objectId = this.suinsResolver.hardcodedSubdmains(parsedUrl.subdomain);
-        if (!objectId && !parsedUrl.subdomain.includes(".")) {
+    	// Resolve to an objectId using a hard-coded subdomain.
+    	const hardCodedObjectId = this.suinsResolver.hardcodedSubdomains(parsedUrl.subdomain);
+		if (hardCodedObjectId) return hardCodedObjectId;
+
+        // If b36 subdomains are supported, resolve them by converting them to a hex object id.
+		const isSuiNSDomain = parsedUrl.subdomain.includes(".");
```
giac-mysten, line 129: Important: this is always `false`, and the current code only works because `subdomaintoobjectid` fails if the name is not a suins name

giac-mysten (approved):
This works. There is a problem in the resolution path if `b36DomainResolutionSupport` is `true`, but we can solve this in a follow up. Will upgrade later today

PR #425, opened 2025-03-06
feat(docker): add gh action to publish testnet sp to docker hub

Tzal3x:
A new docker image will be published each time we merge to testnet. 
The workflow can also be triggered manually.  In the future, we will have to do the same for mainnet. 

⚠️ **Warning**! I updated the Dockerfile, redeclaring the ARG and ENV variables after the `FROM base AS release` step. 
If we don't do this, the environment variables do not persist when running the container after `docker build -f portal/docker/server/Dockerfile -t server-portal --no-cache . --build-arg ENABLE_SENTRY=false`. Dockerfile ARGs disappear between FROM statements: https://stackoverflow.com/a/56748289.

TODO 
- [ ] Setup a testnet server portal namespace and a walrus sites repository in docker hub. 
- [ ] Update the environment variable and secrets accordingly.

### DIFF
--- a/.github/workflows/docker-hub-testnet-sp.yml
+++ b/.github/workflows/docker-hub-testnet-sp.yml
@@ -50,6 +50,14 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
             ENABLE_SENTRY=false
+            ENABLE_BLOCKLIST=false
+            ENABLE_ALLOWLIST=false
+            LANDING_PAGE_OID_B36=41qecxqcyzqm8gl0cp2fqd6iq62j0jo5thr39nb0bsg39acnib
+            PREMIUM_RPC_URL_LIST=https://fullnode.testnet.sui.io
+            RPC_URL_LIST=https://fullnode.testnet.sui.io,https://testnet.suiet.app
+            SUINS_CLIENT_NETWORK=testnet
+            AGGREGATOR_URL=https://aggregator.walrus-testnet.walrus.space
+            SITE_PACKAGE=0xc5bebae319fc9d2a9dc858b7484cdbd6ef219decf4662dc81a11dc69bb7a5fa7
           no-cache: true
 
       - name: Generate artifact attestation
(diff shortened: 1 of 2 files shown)


portal/docker/server/Dockerfile
```diff
@@ -79,6 +79,44 @@ COPY --from=install /temp/prod/server/package.json server/package.json
 COPY --from=install /temp/prod/server/.next server/.next
 COPY --from=install /temp/prod/server/public server/public/
 
+# environment variables, can be overridden by docker build --build-arg <arg>=<value>
+ARG ENABLE_ALLOWLIST="false"
```
pei-mysten, line 83: no, we should NOT use ARG/ENV in a runtime image, all env vars can be either passed in by `--env` or `--env-file`.  Having these default values in a runtime images could really mess things up

.github/workflows/docker-hub-testnet-sp.yml
```diff
@@ -0,0 +1,60 @@
...
+      attestations: write
+      id-token: write
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
```
pei-mysten, line 33: tbh i don't feel that good about pushing images from this repo, it should follow our release schedule and be a part of the release pipeline. Plus it's not a good idea to put the creds all over the places

pei-mysten (changes requested):
- in short I don't think we should push to mysten docker hub from here, it should be part of our release pipeline and we just push the built images in an additional destination.
- Also, adding ARG/ENV is a no-go, we should stick to `--env` or `--env-file`, having default env var can really mess things up

.github/workflows/docker-hub-testnet-sp.yml
Tzal3x: Understood 👍 - where should I make a PR for this pipeline then?

Tzal3x:
> in short I don't think we should push to mysten docker hub from here, it should be part of our release pipeline and we just push the built images in an additional destination.

Thank you for the review, closing this PR since we agree to implement this elsewhere.

PR #426, opened 2025-03-07
feat: remove walrus-sites-sw.js from server portal

Tzal3x:
This is done to prevent service worker scripts
to mess with our fresh server portal redeployments.


giac-mysten (changes requested):
Requesting changes just to block merging until we can go. 
There is a problem here, if we host on both `walrus.site` and on another domain

Tzal3x:
Since we migrated to a Bun server, this will no longer work. 

I replaced this PR with the following: https://github.com/MystenLabs/walrus-sites/pull/445

PR #427, opened 2025-03-07
docs: restructure and improve README

Tzal3x:
- Our repo visitors most probably glance through our README first before clicking any link to the docs. It would be wiser to catch their attention with a very high level intro about sites.
- Reorder the file structure puting the user-facing components first.
- Use reference-style links instead of inlin links for a cleaner markdown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,57 @@
 # Walrus Sites!
 
-This repo contains the various components of a working [Walrus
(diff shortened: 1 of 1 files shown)


VLegakis (approved):
Great job. Its a great first draft and a very helpful high level overview of the project.
However I believe there may be business owners who need to write-edit this also. Specifically referring to Portal, Enoki, SuiNS domains, Blobs expiration time, gas costs etc needs to be decided by them I think.

PR #432, opened 2025-03-10
feat(common): decompress chunk by chunk

Tzal3x:
Note: currently we don't use anywhere this function. But we might need to decompress zipped files on the portal side at some point.

### DIFF
--- a/portal/common/lib/decompress_data.ts
+++ b/portal/common/lib/decompress_data.ts
@@ -1,44 +1,81 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import { ungzip, inflate, inflateRaw } from 'pako';
+import { Inflate } from 'pako';
 import logger from './logger';
 
 /**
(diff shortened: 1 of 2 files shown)


portal/common/lib/decompress_data.ts
```diff
@@ -1,44 +1,81 @@
...
-        if (contentEncoding === "plaintext") {
-            return data;
-        }
+	try {
+		if (contentEncoding === "plaintext") {
+			return data;
+		}
 
-        const encodingIsSupported = ["gzip", "deflate", "deflate-raw"].includes(contentEncoding);
+		const encodingIsSupported = ["gzip", "deflate", "deflate-raw"].includes(contentEncoding);
```
nikos-terzo, line 19: Should we log a warning or something if `encodingIsSupported == false`?

nikos-terzo (approved):

portal/common/lib/decompress_data.ts
Tzal3x: We could do this on a higher level function when decompressData returns null. 👍

PR #433, opened 2025-03-11
feat: sew 70 use a cookie monster inside the middleware

Tzal3x:
We should not allow the use of cookies for the time being. 

Bonus: fix domain parsing for localhost development to include `127.0.0.1`. This is useful to test out the portal endpoints with bruno/postman.


portal/server/middleware.ts
```diff
@@ -19,6 +20,7 @@ export function middleware(request: NextRequest) {
     const urlRedirect = new URL('/', request.url)
     const response = NextResponse.rewrite(urlRedirect)
     response.headers.set('x-original-url', urlOriginal)
+    CookieMonster.eatCookies(request, response)
```
nikos-terzo, line 23: Why is the `CookieMonster.eatCookies()` running here (at the end of the function)? Does this mean that we do not need to run it in the `walrus-sites-sw.js`, `/api/healthz` and `alreadyAtRoot` cases?

nikos-terzo (approved):
Just some questions, worth considering whether they bring actionable items.

portal/server/middleware.ts
Tzal3x: Nice find but it wouldn't make a difference because `walrus-sites-sw.js` is just a public static file that we serve and it will be removed very soon. Also /api/healthz is just a special case endpoint where we test our blocklist pod's health. 
`alreadyAtRoot` is the landing page which is also in our control so no harm there done either. 

I can move this to the top anyways to cover all the cases anyway! Thank you!

portal/server/app/route.ts
```diff
@@ -67,7 +67,9 @@ export async function GET(req: NextRequest) {
         }
     }
 
-    const atBaseUrl = portalDomain == url.host.split(":")[0];
+	const hostname = url.host.split(":")[0];
+	const isLocalhost = hostname == '127.0.0.1'
```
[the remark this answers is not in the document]
Tzal3x: `localhost` is already covered in our code in the url parser logic. Regarding the rest, I'm sure it could be a nice addition to also cover. Maybe add this in a following PR.

PR #434, opened 2025-03-12
feat: sew 100 include events upon site creation and update

Tzal3x:
## Summary
1. Include events for Site creation, deletion, etc. 
2. Refactor: create a separate Metadata module.

## Testing

1. Rebuild the site-builder (note: this PR updates the packageId in the sites-config.yaml)
2. Publish the examples/snake.
3. Make a curl request to listen to one of the events introduced in the contract:
```
curl -X POST https://fullnode.testnet.sui.io:443 \
-H "Content-Type: application/json" \
-d '{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "suix_queryEvents",
  "params": [
    {
      "MoveModule": {
        "package": "0x9d055e60cc8012d43c396194395e8f80c54afd083d908bdfbc3c37906f59146e",
        "module": "site",
        "type": "0x9d055e60cc8012d43c396194395e8f80c54afd083d908bdfbc3c37906f59146e::events::SiteCreated"
      }
    },
    null,
    3,
    false
  ]
}'
```

**Note**: The package id used above might have changed if there are more commits added to this PR. Please update it accordingly if you want to test it out.

### DIFF
--- a/move/walrus_site/sources/events.move
+++ b/move/walrus_site/sources/events.move
@@ -3,7 +3,7 @@ module walrus_site::events {
     use sui::event;
     use walrus_site::metadata::Metadata;
 
-    public struct SiteCreated has copy, drop {
+    public struct SiteCreatedEvent has copy, drop {
         site_id: ID,
         site_name: String,
         site_metadata_link: Option<String>,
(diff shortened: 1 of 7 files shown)


move/walrus_site/sources/events.move
```diff
@@ -0,0 +1,51 @@
...
+        site_id: ID,
+    }
+
+    public struct SiteNameUpdateEvent has copy, drop {
+        site_id: ID,
+        old_name: String,
+        new_name: String,
+    }
+
+    public fun emit_site_created(site_id: ID, name: String, metadata: &Metadata) {
```
nikos-terzo, line 26: Maybe we want these to be `public(package) fun`, in order to restrict the event emittance in this package.

nikos-terzo (changes requested):
Event emittance should be limited inside this this package.

move/walrus_site/sources/events.move
Tzal3x: Cool! I didn't know that was possible!

PR #443, opened 2025-03-17
chore: using invalid domain names to represent links

Tzal3x:
We're using [https://*.suiobj](about:blank) and https://blobid.walrus/.* . Should be following [RFC 6761 (section 6.4)](https://www.rfc-editor.org/rfc/rfc6761.html#:~:text=6.4.%20%20Domain%20Name%20Reservation%20Considerations%20for%20%22invalid.%22) and use the `.invalid` suffix instead.


nikos-terzo (approved):
Minor: Add code-comment explaining the reasoning behind `.invalid` usage in service-worker.

PR #448, opened 2025-03-19
fix: Return home button fix

VLegakis:
This returns home to https://walrus.site however we may need to fix that once the domain changes
Most importantly we need to rethink where it will redirect in cases of subdomains (e.g mysite.walrus.site etc)


Tzal3x (approved):
✅

PR #450, opened 2025-03-20
chore: seint 1020 create example env files per network

Tzal3x:
1. Rename `.env.example` -> `.env.testnet.example`
2. Create `.env.mainnet.example` files for both sw and sp.


portal/worker/.env.mainnet.example
Tzal3x, line 1: To be defined upon mainnet launch.

portal/server/.env.mainnet.example
Tzal3x, line 1: To be defined upon mainnet launch.

PR #455, opened 2025-03-24
:bug: handle broken backslash path bug on windows

Shuliyey:
change

  * change backslash `\` to `/` on `new_path` function

## Background
walrus-site can't render `\` url as browser (e.g chrome) auto convert url path `\` to `/`, on windows `site-builder` will upload object as `\` when interacting with `walrus-sites` move contract, therefore the object path will be inserted as `\`

## Solution
replace `\` with `/` in object path, before writing object to Walrus

## References
related to https://discord.com/channels/946098997637042178/1268610159064780973/1352827924260196393

image

when this happens, object with more than one path `\` won't render on walrus.site 

e.g https://discord.com/channels/946098997637042178/1268610159064780973/1353012825592959129

image

* on linux/unix deployment https://ace-walrus.walrus.site/assets/test.txt works
* on windows deployment https://theemptyboxclub.walrus.site/images/logo/logo-word-white.png fails (404)

### DIFF
--- a/move/walrus_site/sources/site.move
+++ b/move/walrus_site/sources/site.move
@@ -114,7 +114,7 @@ module walrus_site::site {
 
     /// Creates a new resource path.
     fun new_path(path: String): ResourcePath {
-        ResourcePath { path }
+        ResourcePath { path.replace('\\', '/') }
     }
 
     /// Updates the name of a site.


move/walrus_site/sources/site.move
```diff
@@ -114,7 +114,7 @@ module walrus_site::site {
 
     /// Creates a new resource path.
     fun new_path(path: String): ResourcePath {
-        ResourcePath { path }
+        ResourcePath { path.replace('\\', '/') }
```
Tzal3x, line 117: Wouldn't this work only for Windows then?

Tzal3x (changes requested):

move/walrus_site/sources/site.move
Shuliyey: yes, the issue is `site-builder` don't work for files under `>1` one directory (of the selected directory to publish) on windows

giac-mysten:
Thanks for the contribution @Shuliyey .
I'm thinking a bit about this, and I think a better way to go about this would be to provide platform-specific functions to convert the OS-level path to the site (HTML-level) path. These could also be gated via compile flags. Wdyt?

Tzal3x:
Closing this PR since it has become stagnant for quite a while.

PR #469, opened 2025-04-01
feat(site-builder): improve passing of context to walrus binary

giac-mysten:
The context is now fully passed to the walrus binary, and the configuration also allows setting the `wallet_env`, such that it is not necessary anymore to select the environment using the `sui` CLI.


Tzal3x (approved):
Awesome, thank you @giac-mysten !

PR #477, opened 2025-04-04
feat(site-builder): add config for wallet address, walrus context, singleton config

giac-mysten:
- adds the option to specify a separate `walrus_context`
- adds the option to specify which `wallet_address` to use
- adds the option to specify a singleton config

### DIFF
--- a/site-builder/src/args.rs
+++ b/site-builder/src/args.rs
@@ -12,7 +12,7 @@ use anyhow::{anyhow, ensure, Result};
 use clap::{Parser, Subcommand};
 use serde::{Deserialize, Serialize};
 use sui_sdk::wallet_context::WalletContext;
-use sui_types::base_types::ObjectID;
+use sui_types::base_types::{ObjectID, SuiAddress};
 
 use crate::{util::load_wallet_context, walrus::output::EpochCount};
 
@@ -36,6 +36,19 @@ pub(crate) struct GeneralArgs {
     /// will be used.
     #[clap(long)]
     pub(crate) wallet_env: Option<String>,
+    /// The address to be used for the Sui wallet.
+    ///
+    /// If not specified, the address specified in the sites-config (under `wallet_address`) will be
+    /// used. If the wallet address is also not specified in the config, the address configured in
+    /// the Sui client will be used.
+    #[clap(long)]
+    pub(crate) wallet_address: Option<SuiAddress>,
+    /// The context that will be passed to the Walrus binary.
+    ///
+    /// If not specified, the Walrus context specified in the sites-config will be
+    /// used. If it is also not specified in the config, no context will be passed.
+    #[clap(long)]
+    pub(crate) walrus_context: Option<String>,
     /// The path or name of the walrus binary.
     ///
     /// The Walrus binary will then be called with this configuration to perform actions on Walrus.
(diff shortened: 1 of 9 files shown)


site-builder/src/args.rs
```diff
@@ -36,6 +36,19 @@ pub(crate) struct GeneralArgs {
...
+    /// used. If the wallet address is also not specified in the config, the address configured in
+    /// the Sui client will be used.
+    #[clap(long)]
+    pub(crate) wallet_address: Option<SuiAddress>,
+    /// The context that will be passed to the Walrus binary.
+    ///
+    /// If not specified, the Walrus context specified in the sites-config will be
+    /// used. If it is also not specified in the config, no context will be passed.
+    #[clap(long)]
+    pub(crate) walrus_context: Option<String>,
```
mlegner, line 51: IIUC, this means that the multi-context config for the site-builder can only be used with a multi-context config for Walrus. Is that understanding correct?

site-builder/src/main.rs
```diff
@@ -83,15 +83,18 @@ async fn run() -> Result<()> {
     tracing::info!("initializing site builder");
 
     let args = Args::parse();
+    tracing::info!(?args, "command line arguments");
```
mlegner, line 86: Nit: This could maybe be a debug log?

site-builder/src/main.rs
```diff
@@ -83,15 +83,18 @@ async fn run() -> Result<()> {
...
             ),
         )?;
+
     tracing::info!(?config_path, "loading sites configuration");
-    let mut config = Config::load_multi_config(config_path, args.context.as_deref())?;
+    let mut config = Config::load_from_multi_config(config_path, args.context.as_deref())?;
+    tracing::debug!(?config, "configuration before merging");
 
     // Merge the configs and the CLI args. Serde default ensures that the `walrus_binary` and
     // `gas_budget` exist.
```
mlegner, line 100: Remark: We should probably do something like this for Walrus too... 🤔

site-builder/src/util.rs
```diff
@@ -202,6 +208,26 @@ pub fn load_wallet_context(path: Option<&Path>, wallet_env: Option<&str>) -> Res
...
+            return Err(anyhow!(
+                "Address '{}' not found in wallet config file '{}'.",
+                target_address,
+                path.display()
+            ));
+        }
+        wallet_context.config.active_address = Some(*target_address);
+        tracing::info!(?target_address, "set the wallet address");
+    } else {
+        tracing::info!("no wallet address provided, using the default one");
```
mlegner, line 228: Nit: Can you add the address to the output?

mlegner (approved):
Thanks a lot for the quick work on this, LGTM.

site-builder/src/args.rs
giac-mysten: If I understand your question correctly, no. 
This means that:
- precedence goes to what is passed to `--walrus-context`
- then, to what is present in `walrus_context` in the config (which is also in the singleton config)
- otherwise, no context.

This is different from the Walrus CLI: the `walrus_context` is present also in the singleton.
See [here](https://github.com/MystenLabs/walrus-sites/pull/477/files#diff-7c8ad711c9193f946d6d3dadb74efc978ea75991dd2e2518cd5b31a4842666b6R8).

The `--context` for the site builder is orthogonal, and is never passed to the Walrus CLI.
So i can have `--context`, and no `walrus_context`, and then the singleton should be fine.

mlegner: Ah, nice. 👍

giac-mysten: Tested, multi-context config for site-builder works with the singleton config on walrus.

PR #484, opened 2025-04-16
fix(site-builder): fix config commented value

giac-mysten:
Enables using the `sitemap` command without specifying the `--walrus-package` flag.


Tzal3x (approved):
Nice! 💯

PR #485, opened 2025-04-24
fix(portal): an issue with parsing spaces in the path

ioannischtz:


portal/common/lib/domain_parsing.ts
```diff
@@ -19,7 +19,14 @@ export function getDomain(url: URL, portalNameLength?: Number): string | null {
...
 */
 export function getSubdomainAndPath(url: URL, portalNameLength?: Number): DomainDetails | null {
-    return splitUrl(url, portalNameLength).details;
+    const splitResult = splitUrl(url, portalNameLength);
+    if (!splitResult.details) {
+        return null;
+    }
+    return {
+        subdomain: splitResult.details.subdomain,
+        path: splitResult.details.path.replace(/%20/g, " "),
```
mlegner, line 28: Suggestion: Use something like [`decodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent) to cover other cases besides spaces as well.

ioannischtz: That makes sense, thank you.

mlegner (commented):
Thanks, @ioannischtz! This LGTM, but my knowledge of TS and this codebase is pretty limited. I'd prefer someone from the WS team to take a brief look as well to make sure I didn't overlook anything.

giac-mysten (approved):
Thanks @ioannischtz ! 🙏

portal/common/lib/domain_parsing.test.ts
```diff
@@ -55,7 +55,11 @@ const getSubdomainAndPathTestCases: [string, DomainDetails][] = [
         { subdomain: 'subsubname.subname.suinsname', path: '/index.html' }],
     ['http://docs.localhost/css/print.css', { subdomain: 'docs', path: '/css/print.css' }],
     ['http://docs.localhost/assets/index-a242f32b.js',
-        { subdomain: 'docs', path: '/assets/index-a242f32b.js'}]
+        { subdomain: 'docs', path: '/assets/index-a242f32b.js'}],
+    ['https://mystenlabs-logos.wal.app/02_Horizontal%20Logo/index.html',
+        { subdomain: 'mystenlabs-logos', path: '/02_Horizontal Logo/index.html' }],
+    ['http://my-site.localhost/files/report%2D2024.pdf',
+        { subdomain: 'my-site', path: '/files/report-2024.pdf' }]
```
giac-mysten, line 62: qq: shouldn't `-` be a valid character? or is this a different unicode version?

Tzal3x (approved):
lgtm~

portal/common/lib/domain_parsing.test.ts
ioannischtz: @giac-mysten Yeah, you're right about that one. Was trying to think of a case other than "space".
Mhh, maybe we could test for the "+" character (%2B). 
Or maybe add a test case with all special characters?

giac-mysten: I guess `-` has two possible representations? can we test both at least?

PR #487, opened 2025-05-13
feat(site-builder): implement the deploy command, that consolidates the publish and update commands

ioannischtz:
- Implement the `Deploy` command, that should consolidate the `Publish` & `Update` commands
- Add the `object_id` field on the `WSResources` struct, to be picked up by the `Deploy` command. When deploying a new-site this field is filled with the actual new object id. On subsequent `Deploy` calls, this field is utilized to perform the update.

### DIFF
--- a/examples/snake/ws-resources.json
+++ b/examples/snake/ws-resources.json
@@ -19,6 +19,6 @@
 	   "project_url": "https://github.com/MystenLabs/walrus-sites/",
 	   "creator": "MystenLabs"
 	},
-  "site-name": "Walrus Snake Game",
+  "site_name": "Walrus Snake Game",
   "ignore": ["/private/*", "/secret.txt"]
 }
--- a/site-builder/src/args.rs
+++ b/site-builder/src/args.rs
@@ -224,6 +224,12 @@ pub(crate) enum Commands {
         /// The name of the site.
         #[clap(short, long)]
         site_name: Option<String>,
+        /// The object ID of a partially published site to be completed.
+        ///
+        /// This is the object ID of the site that was published before, and is now being updated.
+        /// If this is provided, it will override the object ID in the ws-resources.json file.
+        #[clap(short, long)]
+        object_id: Option<ObjectID>,
         #[clap(short, long, action)]
         watch: bool,
         #[clap(long, action)]
--- a/site-builder/src/args.rs
+++ b/site-builder/src/args.rs
@@ -213,7 +213,11 @@ impl ObjectIdOrName {
 #[clap(rename_all = "kebab-case")]
 pub(crate) enum Commands {
     /// Deploy a new site on Sui.
-    /// Combines the `publish` and `update` commands.
+    ///
+    /// If the site has not been published before, this command publishes it and stores
+    /// the object ID of the Site in the ws-resources.json file.
+    /// If the site has been published before, this command updates the site(indicaded
+    /// by the site_object_id field in the ws-resources.json file).
     Deploy {
         #[clap(flatten)]
         publish_options: PublishOptions,
@@ -226,6 +230,7 @@ pub(crate) enum Commands {
         check_extend: bool,
     },
     /// Publish a new site on Sui.
+    #[deprecated(note = "This command is soon to be deprecated; please use `deploy` instead")]
     Publish {
         #[clap(flatten)]
         publish_options: PublishOptions,
@@ -234,6 +239,7 @@ pub(crate) enum Commands {
         site_name: Option<String>,
     },
     /// Update an existing site.
+    #[deprecated(note = "This command is soon to be deprecated; please use `deploy` instead")]
     Update {
         #[clap(flatten)]
         publish_options: PublishOptions,
--- a/site-builder/src/args.rs
+++ b/site-builder/src/args.rs
@@ -212,6 +212,19 @@ impl ObjectIdOrName {
 #[derive(Subcommand, Debug)]
 #[clap(rename_all = "kebab-case")]
 pub(crate) enum Commands {
+    /// Deploy a new site on Sui.
+    /// Combines the `publish` and `update` commands.
+    Deploy {
+        #[clap(flatten)]
+        publish_options: PublishOptions,
+        /// The name of the site.
+        #[clap(short, long)]
+        site_name: Option<String>,
+        #[clap(short, long, action)]
+        watch: bool,
+        #[clap(long, action)]
+        check_extend: bool,
+    },
     /// Publish a new site on Sui.
     Publish {
         #[clap(flatten)]
--- a/site-builder/src/args.rs
+++ b/site-builder/src/args.rs
@@ -236,7 +236,6 @@ pub(crate) enum Commands {
         check_extend: bool,
     },
     /// Publish a new site on Sui.
-    #[deprecated(note = "This command is soon to be deprecated; please use `deploy` instead")]
     Publish {
         #[clap(flatten)]
         publish_options: PublishOptions,
@@ -245,7 +244,6 @@ pub(crate) enum Commands {
         site_name: Option<String>,
     },
     /// Update an existing site.
-    #[deprecated(note = "This command is soon to be deprecated; please use `deploy` instead")]
     Update {
         #[clap(flatten)]
         publish_options: PublishOptions,
--- a/site-builder/src/args.rs
+++ b/site-builder/src/args.rs
@@ -230,8 +230,27 @@ pub(crate) enum Commands {
         /// If this is provided, it will override the object ID in the ws-resources.json file.
         #[clap(short, long)]
         object_id: Option<ObjectID>,
+        /// Watch the site directory for changes and automatically redeploy when files are modified.
+        ///
+        /// When enabled, the command will continuously monitor the site directory and trigger a
+        /// redepoloyment whenever changes are detected, allowing for rapid development iteration.
         #[clap(short, long, action)]
         watch: bool,
+        /// Checks and extends all blobs in an existing site during an update.
+        ///
+        /// With this flag, the site-builder will force a check of the status of all the Walrus
+        /// blobs composing the site, and will extend the ones that expire before `--epochs` epochs.
+        /// This is useful to ensure all the resources in the site are available for the same
+        /// amount of epochs.
+        ///
+        /// Further, when this flag is set, _missing_ blobs will also be reuploaded (e.g., in case
+        /// they were deleted, or are all expired and were not owned, or, in case of testnet, the
+        /// network was wiped).
+        ///
+        /// Without this flag, when updating a site, the `deploy` command will only create new blobs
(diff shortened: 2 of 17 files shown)


site-builder/src/main.rs
```diff
@@ -103,6 +103,58 @@ async fn run() -> Result<()> {
...
+                            publish_options,
+                            SiteIdentifier::NewSite(site_name),
+                            ContinuousEditing::Once,
+                            BlobManagementOptions::no_status_check(),
+                        )
+                        .run()
+                        .await?
+                }
+            }
+        }
```
giac-mysten, line 157: Suggestion: this can be compressed furthere, something like the following

```suggestion
            // Load the ws-resources file, to check for the site-object-id. If it exists, it means
            // the site is already deployed, in which case we should do the same as the update
            // command. If it doesn't exist, we can deploy the site, by doing the same as the
            // publish command.
            let (ws_resources, _) = load_ws_resources(
                publish_options.walrus_options.ws_resources.as_deref(),
                publish_options.directory.as_path(),
            )?;
            let site_object_id = ws_resources.as_ref().and_then(|res| res.site_object_id);

            let (identifier, continuous_editing, blob_management) = match site_object_id {
                Some(object_id) => (
                    SiteIdentifier::ExistingSite(object_id),
                    ContinuousEditing::from_watch_flag(watch),
                    BlobManagementOptions { check_extend },
                ),
                None => (
                    SiteIdentifier::NewSite(site_name.unwrap_or_else(|| {
                        ws_resources
                            .and_then(|res| res.site_name)
                            .unwrap_or_else(|| "Test Site".to_string())
                    })),
                    ContinuousEditing::Once,
                    BlobManagementOptions::no_status_check(),
                ),
            };

            SiteEditor::new(args.context, config)
                .with_edit_options(
                    publish_options,
                    identifier,
                    continuous_editing,
                    blob_management,
                )
                .run()
                .await?
```

site-builder/src/publish.rs
```diff
@@ -270,6 +270,57 @@ impl SiteEditor<EditOptions> {
         )
         .await?;
         let (response, summary) = site_manager.update_site(&local_site_data).await?;
+
+        // -- Save the site_object_id to the ws-resources file.
```
giac-mysten, line 274: Nit:
```suggestion
        // Save the site_object_id to the ws-resources file.
```

giac-mysten: Suggestion: if you need a header for code, better separate what comes next into a function :)

site-builder/src/publish.rs
```diff
@@ -270,6 +270,57 @@ impl SiteEditor<EditOptions> {
...
+            Err(e) => {
+                // If we can't get the ID, we can't save it.
+                display::error(format!(
+                    "Could not get site object ID from transaction response: {}",
+                    e
+                ));
+                // Maybe return early here ?? :
+                // return Err(e.context("Failed to extract site object ID for saving"));
+                return Ok((site_manager.active_address()?, response, summary)); // Exit OK without saving
+            }
```
giac-mysten, line 302: Remark: Good point, i've been wanting to improve the `get_site_id_from_response` path for a while now. The errors are too opaque most of the time. I'll add a linear issue for this.

Suggestion: 
- remove the `#[allow(dead_code)]` from `get_site_id_from_response` 
- have `get_site_id_from_response` return `ObjectID`, without the `Result`
- add to the documentation of the function the `# Panics` reason
- remove this `match` entirely

site-builder/src/publish.rs
```diff
@@ -270,6 +270,57 @@ impl SiteEditor<EditOptions> {
...
+        };
+        match ws_resources {
+            Some(mut ws_resources) => {
+                ws_resources.site_object_id = Some(new_site_object_id);
+                ws_resources.save(ws_resources_path.unwrap())?;
```
giac-mysten, line 307: Important: `ws_resources_path` is not guaranteed to exist. This will panic if the user does not specify a path and there is no ws-resource file at the default location.

Suggestion: something like
```
                    ws_resources_path
                        .unwrap_or_else(|| self.directory().join(DEFAULT_WS_RESOURCES_FILE)),

```

site-builder/src/publish.rs
```diff
@@ -270,6 +270,57 @@ impl SiteEditor<EditOptions> {
...
+            }
+            None => {
+                // TODO: Maybe implement the `Default` trait for `WSResources` ?
+                // If the ws_resources file does not exist, create it with the new site object ID.
```
giac-mysten, line 311: Remark: yes, this is a good point.
Then you can simplify the following by 

```
WSResources {
  site_object_id: Some(site_object_id),
  ..Default::default()
}

File: site-builder/src/publish.rs [hunk 6]
```diff
@@ -270,6 +270,57 @@ impl SiteEditor<EditOptions> {
...
+                let ws_resources = WSResources {
+                    headers: None,
+                    routes: None,
+                    metadata: None,
+                    site_name: None,
+                    site_object_id: Some(new_site_object_id),
+                    ignore: None,
+                };
+                ws_resources.save(ws_resources_path.unwrap())?;
```
giac-mysten, line 320: Important: same as above.

site-builder/src/site/config.rs
```diff
@@ -41,6 +45,17 @@ impl WSResources {
         tracing::info!(?ws_config, "ws resources configuration loaded");
         Ok(ws_config)
     }
+
+    /// Saves the `WSResources` struct to a json file, pretty-printed.
+    pub fn save<P: AsRef<Path>>(&self, path: P) -> Result<()> {
+        tracing::info!(file=%path.as_ref().display(), "saving Walrus site resources");
+        let file = std::fs::File::create(path.as_ref())
+            .with_context(|| format!("Failed to create file: {}", path.as_ref().display()))?;
```
giac-mysten, line 53: Suggestion: Since the `Result` from the `anyhow` crate, use `anyhow::Context` as above instead of `core::result::Result.with_context`.

giac-mysten (commented):
Thanks @ioannischtz ! this already looks good! a few comments from my side.

Tzal3x:
Question: Should we tag publish and update as deprecated commands? They can still be available for backwards compatibility but we should encourage devs to prefer the deploy command.

site-builder/src/main.rs
ioannischtz: Nice refactor suggestion @giac-mysten , thank you!

site-builder/src/publish.rs
```diff
@@ -357,6 +397,58 @@ fn print_summary(
...
+
+    // If not already set, try to get the site object ID from the transaction effects.
+    display::action("Getting the site object ID from the transaction effects");
+
+    // potentially panics if the site object ID is not found.
+    let new_site_object_id = get_site_id_from_response(active_address, tx_effects);
+    // If we reach here, get_site_id_from_response succeeded.
+
+    // Update the existing ws_resources or create a new one, then save.
+    match current_ws_resources {
```
giac-mysten, line 425: Suggestion: when unpacking an `Option`, use `if let Some` instead of `match` (a bit more compact)

site-builder/src/publish.rs
```diff
@@ -357,6 +397,58 @@ fn print_summary(
...
+        Some(mut ws_res) => {
+            ws_res.site_object_id = Some(new_site_object_id);
+            display::action(format!(
```
giac-mysten, line 428: Suggestion: since you print the result of this operation later as well, I'd remove it from there (and below as well)

giac-mysten: alternatively, you could remove the `SiteIdSaveResult`, and keep the print only here.

giac-mysten (commented):
Thanks @ioannischtz , another round of comments, but this looks great!

examples/snake/ws-resources.json
```diff
@@ -19,6 +19,6 @@
 	   "project_url": "https://github.com/MystenLabs/walrus-sites/",
 	   "creator": "MystenLabs"
 	},
-  "site-name": "Walrus Snake Game",
+  "site_name": "Walrus Snake Game",
```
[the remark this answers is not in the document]
ioannischtz: Oh, this is from another PR. (I rebased on top of main).
yes, that was for consistency, because some users were getting confused on Discord.

Tzal3x (approved):
Thank you @ioannischtz !

site-builder/src/args.rs
```diff
@@ -212,6 +212,29 @@ impl ObjectIdOrName {
...
+        site_name: Option<String>,
+        /// The object ID of a partially published site to be completed.
+        ///
+        /// This is the object ID of the site that was published before, and is now being updated.
+        /// If this is provided, it will override the object ID in the ws-resources.json file.
+        #[clap(short, long)]
+        object_id: Option<ObjectID>,
+        #[clap(short, long, action)]
+        watch: bool,
+        #[clap(long, action)]
```
nikos-terzo, line 235: Minor: Why not copy the docstring from `check_extend` in update below?
It would also be nice to have a docstring for `watch` as well.

ioannischtz: For sure, I'll add it.

site-builder/src/util.rs
```diff
@@ -109,16 +113,23 @@ pub fn id_to_base36(id: &ObjectID) -> Result<String> {
...
     effects: &SuiTransactionBlockEffects,
-) -> Result<ObjectID> {
+) -> ObjectID {
+    // Return type changed to ObjectID
     tracing::debug!(
         ?effects,
         "getting the object ID of the created Walrus site."
     );
-    Ok(effects
+    effects
```
nikos-terzo, line 132: Not really in scope of this PR, but food for thought:
It doesn't seem ok to get the site-id from using the active-address and looking for objects created under this owner. It assumes a lot of things:
1. Only one object created - the site.
2. The site is returned to the current active address.
3. The active address doesn't change.
I think we should redesign this to use object-changes instead and first verify any assumptions we do.

ioannischtz: Sounds good, let's open a new ticket/PR for this.

site-builder/src/util.rs
```diff
@@ -237,6 +248,80 @@ pub fn load_wallet_context(
...
+    site_object_id: ObjectID,
+    site_name: Option<String>,
+    initial_ws_resources_opt: Option<WSResources>,
+    ws_resources_path: &Path,
+) -> Result<WSResources, anyhow::Error> {
+    let mut ws_resources_to_save = initial_ws_resources_opt.unwrap_or_default();
+
+    // Update/Set the site_object_id
+    if ws_resources_to_save.site_object_id != Some(site_object_id) {
+        tracing::info!(
```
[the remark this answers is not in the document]
ioannischtz: Mhh, yeah debug probably better.

nikos-terzo (approved):
Only some minor comments

PR #488, opened 2025-05-13
fix(site-builder): Updating an existing site should also update its metadata in case of change.

nikos-terzo:


site-builder/src/site/manager.rs
```diff
@@ -305,7 +305,16 @@ impl SiteManager {
...
         let mut ptb = match &self.site_id {
             SiteIdentifier::ExistingSite(site_id) => {
-                ptb.with_call_arg(&self.wallet.get_object_ref(*site_id).await?.into())?
+                let ptb = ptb.with_call_arg(&self.wallet.get_object_ref(*site_id).await?.into())?;
+                // Also update metadata if there is a diff
+                match updates.metadata_op {
+                    MetadataOp::Update => {
+                        // REVIEW: If metadata inside ws-resources, is None (Metadata::default())
+                        // and different from the site published, should we still update?
+                        ptb.with_update_metadata(self.metadata.clone().unwrap_or_default())?
```
Tzal3x, line 314: Good question. I think we should not because the implied behaviour is that in the absence of a metadata field, no _changes_ should be applied.
The default metadata should be applied only when the site is first published. 
What do you think?

Tzal3x (approved):
Thanks! Left a comment regarding the default metadata.

site-builder/src/site/manager.rs
nikos-terzo: I agree, updated to require `metadata` field in ws-resources existing for `MetadataOp::Update`.

Tzal3x (approved):
Thanks @nikos-terzo !

PR #489, opened 2025-05-15
fix(site-builder): Enforce site_name field and set a better default name

ioannischtz:
- On WSResources change the site-name field(with a dash) to site_name
- On WSResources add the #[serde(deny_unknown_fields)] so that we throw an error. 
- On main.rs change the default site name from 'Test Site' to 'My Walrus Site'
- On config.rs add a test case for the invalid/valid site_name field.


Tzal3x (approved):
lgtm 🔥

PR #500, opened 2025-06-16
chore(portal): sew 234 add versions to sw and sp portals

Tzal3x:
#### Service Worker
The service worker and the server portal should each have a version associated with it.
We don't host a service worker, and we don't put much effort into it to be regarded as production ready, so I set it to 0.1.0.

#### Server Portal
The server portal on the other hand is production ready and battle tested, so I set it to 1.0.0.

### DIFF
--- a/portal/server/package.json
+++ b/portal/server/package.json
@@ -1,6 +1,7 @@
 {
     "name": "server",
-    "version": "0.1.0",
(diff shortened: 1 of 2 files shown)


ioannischtz (approved):
🚀

PR #501, opened 2025-06-20
feat: Add github action for site deploy

teohaik:
Introduces a github action that can be reused by anyone publicly in their own repo. 
The action builds a pnpm-based website and deploys it to walrus as a walrus site

Prerequisites:

- A wallet with SUI and WAL should be passed
- Wallet key in the form of keystore should be passed as secret variable

I've create a second repo with a hello-site as an example that uses the new action defined in our repo for testing things out

- This is exactly how community builders will invoke the site-builder action to publish their site

Tested on examples directory using `gh workflow run`.
- snake: publish + update
- viter-react-ts: publish

### DIFF
--- a/.github/workflows/deploy-snake.yml
+++ b/.github/workflows/deploy-snake.yml
@@ -25,6 +25,6 @@ jobs:
           SUI_NETWORK: ${{ inputs.SUI_NETWORK }}
           SUI_ADDRESS: ${{ vars.SUI_ADDRESS }}
           SUI_KEYSTORE: ${{ secrets.SUI_KEYSTORE }}
-          DIST_FOLDER: examples/snake
+          DIST: examples/snake
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           EPOCHS: ${{ inputs.EPOCHS }}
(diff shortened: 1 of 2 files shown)


.github/actions/publish/action.yml
```diff
@@ -0,0 +1,65 @@
...
+          uses: actions/checkout@v4 # pin@v4
+        - name: Setup pnpm
+          uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 # pin@v4
+          with:
+              version: 9.9.0
+        - name: Setup node
+          uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # pin@v4
+          with:
+              node-version: 22
+        - name: Install dependencies
```
nikos-terzo, line 37: Not everyone generates the static site using node/pnpm.
One could use python, or rust or any other tool to build static sites.
Can we investigate whether there is a way to make the site-build-process generic?

.github/actions/deploy/action.yml
```diff
@@ -0,0 +1,145 @@
...
+        description: |
+          The GITHUB_TOKEN for creating a PR, in case of ws-resources changed during
+          site-publishing.
+          If not provided, the PR creation step will be skipped.
+          When using this, the workflow must have `contents: write` and `pull-requests: write`
+          permissions.
+        required: false
+
+runs:
+    using: "composite"
```
teohaik, line 65: The steps here need commenting so that their actions can be easily understood.

nikos-terzo: Thanks, added!

PR #512, opened 2025-07-04
fix(site-builder): Fix blob_id calculation causing rate limits

ioannischtz:
Looking at the [implementation](https://github.com/MystenLabs/walrus/blob/main/crates/walrus-service/src/client/cli/runner.rs#L1035) of the walrus blob_id method,
it seems that calls to the rpc client are  happening if we do not provide the n_shards or the encoding_type. 

We [do](https://github.com/MystenLabs/walrus-sites/blob/main/site-builder/src/site/resource.rs#L424) provide the n_shards, but not the encoding_type. 

However, as you can see here: [get_sui_read_client_from_rpc_node_or_wallet](https://github.com/MystenLabs/walrus/blob/429d070f72d9fb20e7a8a20fc0dd757938f66bae/crates/walrus-service/src/client/cli/runner.rs#L1049)  a DEFAULT_ENCODING constant is used, so we could potentially open a small PR on walrus as well, so that it doesn't try to setup an rpc client in this case.

### DIFF
--- a/examples/gen_files.sh
+++ b/examples/gen_files.sh
@@ -1,0 +1,132 @@
+# Copyright (c) Mysten Labs, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+#!/usr/bin/env bash
+
+# Utilize this script to generate a folder with random files,
+# so we can test deployment of a site with multiple files
+
+set -euo pipefail
+shopt -s nullglob
+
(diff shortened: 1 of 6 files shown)


ioannischtz:
Maybe I should add an e2e test for this case? (200+ files)

PR #515, opened 2025-07-16
feat(ci/cd): add workflow to build and publish server-portal Docker image

Tzal3x:
This workflow actually triggers a gh action to another repository that actually does the work.


nikos-terzo (approved):
LGTM

PR #521, opened 2025-07-21
fix(gh-action): tarball creation step

Tzal3x:
This is needed so that we can avoid the "file changed as we read it" error.


nikos-terzo (approved):
Seems good, but it might be cleaner to just store the tar-file in a different folder rather than copying everything somewhere else.

PR #522, opened 2025-07-25
feat(site-builder): Support json mode in site-builder input

nikos-terzo:


site-builder/src/args.rs
```diff
@@ -234,17 +259,51 @@ impl ObjectIdOrName {
...
+        ///     "publish": {
+        ///       "publishOptions": {
+        ///         "directory": "/path/to/site_directory",
+        ///         "walrusOptions": {
+        ///           "epochs": 1
+        ///         }
+        ///       }
+        ///     }
+        ///   }
+        /// }
```
Tzal3x, line 293: This is outdated, I made the site-builder work by using this:
```json
{
    "config":"sites-config.yaml",
    "command":{
        "publish":{
            "directory":"examples/snake",
            "epochs":1
        }
    },
    "context": "testnet"
}
```

site-builder/tests/publish.rs
```diff
@@ -38,3 +38,34 @@ async fn publish_snake() -> anyhow::Result<()> {
...
+            command_string: Some(format!(
+                r#"{{
+            "config":"{}",
+            "command":{{
+                "publish":{{
+                    "directory":"{}",
+                    "epochs":1
+                }}
+            }}
+        }}"#,
```
Tzal3x, line 62:
```suggestion
                r#"{{
            "config":"{}",
            "context": "{}" // added context 
            // ... add more here
            "command":{{
                "publish":{{
                    "directory":"{}",
                    "epochs":1
                }}
            }}
        }}"#,
```
suggestion: let's define explicitly as many options as possible so it can act as a reference for devs.

Tzal3x (changes requested):
Overall looks good. I tested it out with the following `tmp.json`:
```json
{
    "config":"sites-config.yaml",
    "command":{
        "publish":{
            "directory":"examples/snake",
            "epochs":1
        }
    },
    "context": "testnet"
}
```
And then:

`site-builder --context=testnet json < tmp.json`

It worked! :ok_man:

Tzal3x (approved):
Thank you @nikos-terzo !

site-builder/tests/publish.rs
nikos-terzo: I can add it to the Json docstring though.

site-builder/src/args.rs
```diff
@@ -23,86 +23,130 @@ use crate::{
...
     pub context: Option<String>,
     #[clap(flatten)]
+    #[serde(default, flatten)]
     pub general: GeneralArgs,
     #[command(subcommand)]
     pub command: Commands,
 }
 
+impl Args {
+    pub fn extract_json_command(self) -> Result<Self> {
```
giac-mysten, line 47: Question: in the Walrus CLI we do allow for nested JSON commands (i.e., mutliple layers of JSON). I think this is not the case here, correct?

site-builder/src/args.rs
```diff
@@ -23,86 +23,130 @@ use crate::{
...
+        // too. This is used to support the same behavior in json.
+        let general_inside_command = hoist_general_args(command_string)?;
+        new_self.general.merge(&general_inside_command);
+
+        // new_self.json = true;
+        Ok(new_self)
+    }
+}
+
+fn hoist_general_args(serialized: &str) -> Result<GeneralArgs> {
```
giac-mysten, line 75: I'm not opposed to this :) will need to think a bit if there are other ways.

giac-mysten (commented):
A first round of comments. Thanks, @nikos-terzo !

site-builder/tests/publish.rs
```diff
@@ -38,3 +38,39 @@ async fn publish_snake() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn json_publish_snake() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
nikos-terzo: Currently we are not using `test_` prefix for integration tests, as they are under the tests directory, and I think this is explanatory enough.

site-builder/src/args.rs
```diff
@@ -23,86 +23,130 @@ use crate::{
...
+                &std::io::read_to_string(std::io::stdin())?
+            }
+        };
+        tracing::debug!(
+            command = command_string.replace('\n', ""),
+            "running JSON command"
+        );
+        let mut new_self: Args = serde_json::from_str(command_string)?;
+
+        // Someone might pass a global-argument inside the command in json, as clap allows it
```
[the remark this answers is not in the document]
nikos-terzo: Thanks! Committed.

site-builder/src/args.rs
```diff
@@ -234,17 +278,51 @@ impl ObjectIdOrName {
...
+    Json {
+        /// The JSON-encoded args for the Walrus CLI; if not present, the args are read from stdin.
+        ///
+        /// The JSON structure follows the CLI arguments, containing global options and a "command"
+        /// object at the root level. The "command" object itself contains the command (e.g.,
+        /// "publish", "update", "deploy", "convert", ...) with an object containing the command
+        /// options.
+        ///
+        /// Note that where CLI options are in "kebab-case", the respective JSON strings are in
+        /// "camelCase".
```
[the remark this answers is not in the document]
nikos-terzo: Thanks! Committed!

site-builder/src/args.rs
nikos-terzo: Do you maybe mean [this part](https://github.com/MystenLabs/walrus/blob/main/crates/walrus-service/src/client/cli/args.rs#L168-L175)? I do not see how this is different, as these are flattened.

giac-mysten: I mean [this part](https://github.com/MystenLabs/walrus/blob/5816a7103260acc6f6b60d28dbff5c98afe79545/crates/walrus-service/src/client/cli/args.rs#L104-L124)

site-builder/src/args.rs
```diff
@@ -23,86 +23,130 @@ use crate::{
...
     /// The path to the configuration file for the site builder.
-    #[arg(short, long)]
+    #[arg(short, long, global = true)]
     pub config: Option<PathBuf>,
     /// The context with which to load the configuration.
     ///
     /// If specified, the context will be taken from the config file. Otherwise, the default
     /// context, which is also specified in the config file, will be used.
-    #[arg(long)]
+    #[arg(long, global = true)]
```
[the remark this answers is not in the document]
giac-mysten: but there is no difference from a user UX perspective, correct?

site-builder/src/args.rs
```diff
@@ -23,86 +23,130 @@ use crate::{
...
+            }
+        };
+        tracing::debug!(
+            command = command_string.replace('\n', ""),
+            "running JSON command"
+        );
+        let mut new_self: Args = serde_json::from_str(command_string)?;
+
+        // Someone might pass a global-argument inside the command in json, as clap allows them
+        // too. This is used to support the same behavior in json.
```
giac-mysten, line 66:
```suggestion
        // to. This is used to support the same behavior in json.
```

site-builder/src/args.rs
giac-mysten, line 47: Suggestion: Could we add tests for this method, that provide the same commeand with the args in different positions, and check that the resulting struct is the same?

site-builder/src/args.rs
```diff
@@ -23,86 +23,130 @@ use crate::{
...
+        let general_inside_command = Self::hoist_general_args(command_string)?;
+        new_self.general.merge(&general_inside_command);
+
+        // new_self.json = true;
```
giac-mysten, line 70: Nit: remove this? Or do we need to switch the flag?

site-builder/src/args.rs
```diff
@@ -23,86 +23,130 @@ use crate::{
...
+        Ok(new_self)
+    }
+
+    fn hoist_general_args(args_json: &str) -> Result<GeneralArgs> {
```
giac-mysten, line 74: Thought a bit about how to make this more readable/explicit. The only thing I could come up with is to create a helper struct for deserialization, looking like:

```rust
      #[derive(Deserialize)]
      #[serde(untagged)]
      enum CommandWrapper {
          Publish {
              #[serde(flatten)]
              general: GeneralArgs,
              #[serde(flatten)]
              _rest: serde_json::Value,
          },
          Update { ...
```

but this would be a pain to maintain. I think what you have is good!

site-builder/src/args.rs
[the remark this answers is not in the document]
nikos-terzo: I think, there wasn't any to begin with, was there?

site-builder/src/args.rs
nikos-terzo: Thank you for looking into this anyway.
I think that I saw somewhere a feature request for serde supporting `#[serde(global)]`, but I may be imagining things (cannot find it now). In other words, it would be great to have someone else maintain it.
However, given the current limitations, I will stick with the function, and a better docstring as above.

site-builder/src/args.rs
nikos-terzo: I just thought of this: What happens if the user passes something like:
```
site-builder json '{"command": {"json":{"command_str": ... }}}
```
I just tried this with walrus and it hangs 😅
```
walrus json '{"command":{"json":{"command_str":{"context":"testnet","gasBudget":500000000,"command":{"info":{}}}}}}'
```

site-builder/src/args.rs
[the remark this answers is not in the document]
nikos-terzo: Actually, I now see your point. If we pass `context` and `config` inside `GeneralArgs` we essentially make them `global` for json mode too, correct? Will do.

site-builder/src/args.rs
nikos-terzo: We also loop now over nested json commands.
The only difference is that we assign to `tmp_self` instead of `std::mem::replace`.

PR #525, opened 2025-07-29
ci: Deploy action, support wallet exported keys

nikos-terzo:
Support bech32 suiprivkey and secret-phrase private keys as exported from Slush wallet.

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -166,14 +166,17 @@ runs:
             echo "📦 Installing suiup to manage Sui CLI tools..."
             curl -sSf https://raw.githubusercontent.com/MystenLabs/suiup/main/install.sh | bash
             
-            # TODO: Test if install.sh handles PATH automatically
-            # Add suiup to PATH for current and subsequent steps
-            # echo "$HOME/.local/bin" >> "$GITHUB_PATH"
-            
-            # Verify installation works (test if PATH is set by install.sh)
-            # export PATH="$HOME/.local/bin:$PATH"
+            # Verify installation works
             suiup --version
             echo "✅ Suiup installed successfully"
+            
+            # Install Sui CLI for the specified network
+            echo "📥 Installing Sui CLI for ${{ inputs.SUI_NETWORK }} network..."
+            suiup install sui@${{ inputs.SUI_NETWORK }} -y
+            
+            # Verify Sui CLI installation
+            sui --version
+            echo "✅ Sui CLI installed successfully"
 
         # Setup Sui account (handles both direct inputs and derivation)
         - name: Setup Sui Account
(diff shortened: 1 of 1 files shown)


ioannischtz (approved):
Left a comment.
Other than that, everything seems fine.

.github/actions/deploy/action.yml
```diff
@@ -56,6 +94,121 @@ inputs:
...
+              import_result=$(sui keytool import "$KEY_STRING" ed25519 --json)
+              
+              # Extract address from the JSON result
+              sui_address=$(echo "$import_result" | jq -r '.suiAddress')
+              
+              # Get the keystore content (this will be in ~/.sui/sui_config/sui.keystore)
+              keystore_content=$(cat ~/.sui/sui_config/sui.keystore)
+              
+              echo "✅ Sui account derived successfully"
+              echo "📍 Address: $sui_address"
```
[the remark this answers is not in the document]
nikos-terzo: `site-builder`, already echoes the site-object-id, thus checking the address is a very simple step. I don't think we need to hide it.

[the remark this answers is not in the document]
ioannischtz: I hadn't thought about that.
yeah, in that case, I'm with you

.github/actions/deploy/action.yml
```diff
@@ -62,6 +100,121 @@ inputs:
...
     steps:
+        # Validate deployer Sui account setup inputs and determine which method is being used
+        - name: Validate Sui Account Setup
+          id: sui_auth_method
+          shell: bash
+          env:
+            SUI_ADDRESS: ${{ inputs.SUI_ADDRESS }}
+            SUI_KEYSTORE: ${{ inputs.SUI_KEYSTORE }}
+            SUI_BECH32_PRIVATE_KEY: ${{ inputs.SUI_BECH32_PRIVATE_KEY }}
+            SUI_SECRET_PHRASE: ${{ inputs.SUI_SECRET_PHRASE }}
```
rjs-mysten, line 111: I don't like this being in inputs, I'm going to think about whether we can do this in a top level environment (i.e. as actual secrets). The fact that the keys are here means they need to be shared somehow, which I also don't like.

rjs-mysten: talked to @nikos-terzo offline, the context of this is to be used by third parties, so this works. however, we need to make sure that the actions README instructs users to use secret environments in order to pass in the input, and do not pass their own inputs directly - https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets

rjs-mysten (approved):

rjs-mysten (approved):

.github/actions/deploy/action.yml
nikos-terzo: Thank you @rjs-mysten. Will create an own repo for this action with appropriate README in a separate task.

PR #526, opened 2025-07-29
ci: Optionally use --check-extend on site deploy action.

nikos-terzo:


ioannischtz (approved):
Everything seems fine, gj.

PR #529, opened 2025-08-06
fix(site-builder): Update the site name if changed in ws-resources.json

Tzal3x:
When  we updated the site name in the `ws-resources.json`, and then ran `site-builder update ...` without any files being modified, then no changes would occur. This is incorrect, the site name should be updated as it is done with the site metadata.

### DIFF
--- a/site-builder/src/publish.rs
+++ b/site-builder/src/publish.rs
@@ -268,10 +268,7 @@ impl SiteEditor<EditOptions> {
             None => None,
         };
 
-        let site_name = match ws_resources.clone() {
-            Some(value) => value.site_name,
-            None => None,
-        };
+        let site_name = ws_resources.as_ref().and_then(|r| r.site_name.clone());
 
         let mut site_manager = SiteManager::new(
             self.config.clone(),
--- a/site-builder/src/site.rs
+++ b/site-builder/src/site.rs
@@ -164,7 +164,7 @@ impl SiteData {
     }
 
     fn site_name_diff(&self, start: &Self) -> SiteNameOp {
-        if self.site_name != start.site_name {
+        if self.site_name.is_some() && self.site_name != start.site_name {
             SiteNameOp::Update
         } else {
             SiteNameOp::Noop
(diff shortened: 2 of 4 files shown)


site-builder/src/site/manager.rs
```diff
@@ -325,6 +328,15 @@ impl SiteManager {
             }
         };
 
+        if let Some(site_name) = &self.site_name {
```
nikos-terzo, line 331: Minor, I think this might be more readable:
```
        if let (Some(site_name), SiteNameOp::Update) = (&self.site_name, updates.site_name_op) {
            ptb.update_name(site_name)?;
        }
```

site-builder/src/publish.rs
```diff
@@ -267,12 +268,18 @@ impl SiteEditor<EditOptions> {
             None => None,
         };
 
+        let site_name = match ws_resources.clone() {
```
nikos-terzo, line 271: Here we shouldn't clone the whole `ws_resources`. Instead we can do something like:
```
let site_name = match ws_resources.as_ref().map(|v| v.site_name.clone());
```

site-builder/src/site.rs
```diff
@@ -149,6 +163,14 @@ impl SiteData {
         }
     }
 
+    fn site_name_diff(&self, start: &Self) -> SiteNameOp {
+        if self.site_name != start.site_name {
```
nikos-terzo, line 167: Here shouldn't we also check that `self.site_name.is_some()` (similar to `metadata_diff`)?

site-builder/src/types.rs
```diff
@@ -274,6 +274,18 @@ impl MetadataOp {
     }
 }
 
+#[derive(Debug, Clone)]
```
nikos-terzo, line 277: Maybe we can give `Copy` to such simple enums too? Relevant to:
```
        if let (Some(site_name), SiteNameOp::Update) = (&self.site_name, updates.site_name_op) {
            ptb.update_name(site_name)?;
        }
```
^ Without Copy, this code will need `&` for the second value in the tuples.

nikos-terzo (changes requested):
Thank you @Tzal3x , good work!
I believe that we should also check that the new site-name is not `None` when setting `SiteNameOp::Update`.
Other than that, only minor/stylistic changes.

nikos-terzo (approved):
LGTM, thanks @Tzal3x !

PR #530, opened 2025-08-07
refactor(site-builder): replace SiteIdentifier with Option<ObjectID>

Tzal3x:
This makes the code more clean since everything gets filtered through the ws-resources.json. 

**Bonus**: Fixes `--site-name` to take precedence over `ws-resources.json > "site-name": ...`.


nikos-terzo (approved):
Only a minor comment on a forgotten mention of `SiteIdentifier` in a docstring.

PR #536, opened 2025-08-26
feat(portal): Retrieve quilt based resources

Tzal3x:
When a site resource is stored as a quilt patch, a different URL should be used to fetch its' contents from the aggregator. 
Therefore, I created two kinds of URL constructors for each case:
1. for blob based resources: `blobAggregatorEndpoint`
2. for quilt based resources: `quiltAggregatorEndpoint`

Bonus:
- Fix the docker image by reintroducing the `@vercel/edge-config` dependency because it was removed by mistake
from `package.json` at commit https://github.com/MystenLabs/walrus-sites/commit/f069cbe59b31c74a0cba33a5a837d09150647770

### DIFF
--- a/portal/common/lib/aggregator.ts
+++ b/portal/common/lib/aggregator.ts
@@ -8,10 +8,8 @@
  * @param aggregatorUrl - The aggregator URL string.
  */
 export function blobAggregatorEndpoint(blob_id: string, aggregatorUrl: string): URL {
-    if (aggregatorUrl.endsWith("/")) {
-        throw new Error("Aggregator URL must not end with a slash.");
-    }
-    return new URL(`${aggregatorUrl}/v1/blobs/${encodeURIComponent(blob_id)}`) as URL;
+	const cleanAggregatorUrl = aggregatorUrl.endsWith("/") ? aggregatorUrl.slice(0, -1) : aggregatorUrl;
+	return new URL(`${cleanAggregatorUrl}/v1/blobs/${encodeURIComponent(blob_id)}`) as URL;
 }
 
 /**
@@ -21,10 +19,8 @@ export function blobAggregatorEndpoint(blob_id: string, aggregatorUrl: string): 
  * @param aggregatorUrl - The aggregator URL string.
  */
 export function quiltAggregatorEndpoint(quilt_patch_id: string, aggregatorUrl: string): URL {
-    if (aggregatorUrl.endsWith("/")) {
-        throw new Error("Aggregator URL must not end with a slash.");
-    }
-    return new URL(
-        `${aggregatorUrl}/v1/blobs/by-quilt-patch-id/${encodeURIComponent(quilt_patch_id)}`,
-    ) as URL;
+	const cleanAggregatorUrl = aggregatorUrl.endsWith("/") ? aggregatorUrl.slice(0, -1) : aggregatorUrl;
+	return new URL(
+		`${cleanAggregatorUrl}/v1/blobs/by-quilt-patch-id/${encodeURIComponent(quilt_patch_id)}`,
+	) as URL;
 }
--- a/portal/common/lib/aggregator.ts
+++ b/portal/common/lib/aggregator.ts
@@ -11,7 +11,7 @@ export function blobAggregatorEndpoint(blob_id: string, aggregatorUrl: string): 
     if (aggregatorUrl.endsWith("/")) {
         throw new Error("Aggregator URL must not end with a slash.");
     }
-    return new URL(`${aggregatorUrl}/v1/blobs/${encodeURIComponent(blob_id)}`) as unknown as URL;
+    return new URL(`${aggregatorUrl}/v1/blobs/${encodeURIComponent(blob_id)}`) as URL;
 }
 
 /**
@@ -26,5 +26,5 @@ export function quiltAggregatorEndpoint(quilt_patch_id: string, aggregatorUrl: s
     }
     return new URL(
         `${aggregatorUrl}/v1/blobs/by-quilt-patch-id/${encodeURIComponent(quilt_patch_id)}`,
-    ) as unknown as URL;
+    ) as URL;
 }
--- a/portal/common/lib/aggregator.ts
+++ b/portal/common/lib/aggregator.ts
@@ -7,9 +7,24 @@
  * @param blob_id - The blob ID to fetch from the aggregator.
  * @param aggregatorUrl - The aggregator URL string.
  */
-export function aggregatorEndpoint(blob_id: string, aggregatorUrl: string): URL {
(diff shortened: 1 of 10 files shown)


portal/common/lib/aggregator.ts
```diff
@@ -7,9 +7,24 @@
...
+ * Returns the URL to fetch the blob by quilt patch ID from the aggregator/cache.
+ *
+ * @param quilt_patch_id - The quilt patch ID to fetch from the aggregator.
+ * @param aggregatorUrl - The aggregator URL string.
+ */
+export function quiltAggregatorEndpoint(quilt_patch_id: string, aggregatorUrl: string): URL {
+    if (aggregatorUrl.endsWith("/")) {
+        throw new Error("Aggregator URL must not end with a slash.");
+    }
+    return new URL(
```
nikos-kitmeridis, line 27: What's the need of adding `as unknown as URL` here?

portal/common/lib/aggregator.ts
```diff
@@ -7,9 +7,24 @@
...
 }
+
+/**
+ * Returns the URL to fetch the blob by quilt patch ID from the aggregator/cache.
+ *
+ * @param quilt_patch_id - The quilt patch ID to fetch from the aggregator.
+ * @param aggregatorUrl - The aggregator URL string.
+ */
+export function quiltAggregatorEndpoint(quilt_patch_id: string, aggregatorUrl: string): URL {
+    if (aggregatorUrl.endsWith("/")) {
```
nikos-kitmeridis, line 24: Why don't we just remove the suffix `/` if it's present instead of throwing?

portal/common/lib/aggregator.ts
```diff
@@ -7,9 +7,24 @@
  * @param blob_id - The blob ID to fetch from the aggregator.
  * @param aggregatorUrl - The aggregator URL string.
  */
```
nikos-kitmeridis, line 9: I had gpt-5 implementing unit tests for this code (mainly to check if type casting was needed or not, see my comment below). 
Can you check them and add them if they look good?
```
import { describe, it, expect } from 'vitest';
import { blobAggregatorEndpoint, quiltAggregatorEndpoint } from './aggregator';

describe('blobAggregatorEndpoint', () => {
  it('builds the correct URL without trailing slash on base', () => {
    const url = blobAggregatorEndpoint('abc123', 'https://agg.example.com');
    expect(url).toBeInstanceOf(URL);
    expect(url.toString()).toBe('https://agg.example.com/v1/blobs/abc123');
  });

  it('throws if aggregatorUrl ends with a slash', () => {
    expect(() => blobAggregatorEndpoint('id', 'https://agg.example.com/')).toThrowError(
      'Aggregator URL must not end with a slash.',
    );
  });

  it('URL-encodes the blob_id', () => {
    const blobId = 'a/b c?d=e&f';
    const url = blobAggregatorEndpoint(blobId, 'https://agg.example.com');
    expect(url.toString()).toBe(
      `https://agg.example.com/v1/blobs/${encodeURIComponent(blobId)}`
    );
  });
});

describe('quiltAggregatorEndpoint', () => {
  it('builds the correct URL without trailing slash on base', () => {
    const url = quiltAggregatorEndpoint('patch-001', 'https://agg.example.com');
    expect(url).toBeInstanceOf(URL);
    expect(url.toString()).toBe(
      'https://agg.example.com/v1/blobs/by-quilt-patch-id/patch-001'
    );
  });

  it('throws if aggregatorUrl ends with a slash', () => {
    expect(() => quiltAggregatorEndpoint('patch-002', 'https://agg.example.com/')).toThrowError(
      'Aggregator URL must not end with a slash.',
    );
  });

  it('URL-encodes the quilt_patch_id', () => {
    const patchId = 'patch id/with?special=chars&and spaces';
    const url = quiltAggregatorEndpoint(patchId, 'https://agg.example.com');
    expect(url.toString()).toBe(
      `https://agg.example.com/v1/blobs/by-quilt-patch-id/${encodeURIComponent(patchId)}`
    );
  });
});
```

portal/common/lib/quilt.ts
```diff
@@ -15,6 +15,11 @@ export class QuiltPatch {
 	* which includes the version, start index, and end index bytes.
 	*/
 	public derive_id(): string {
+		// If quilt_blob_id starts with '0x', remove it.
```
nikos-kitmeridis, line 18: Isn't this comment just noise to the code base?
What's happening in the next 3 lines is pretty obvious.

portal/common/lib/quilt.test.ts
```diff
@@ -1,92 +1,124 @@
...
-
-	it('should handle empty hex string', () => {
-		const buf = QuiltPatch.hexToBuffer('');
+	it("should handle empty hex string", () => {
+		const buf = QuiltPatch.hexToBuffer("");
 		expect(Array.from(new Uint8Array(buf))).toEqual([]);
 	});
 });
 
+describe("extract_internal_identifier", () => {
```
nikos-kitmeridis, line 70: I love it that you added a unit test for your test util function! 💯

portal/common/lib/redirects.ts
```diff
@@ -3,15 +3,17 @@
...
+import { blobAggregatorEndpoint } from "./aggregator";
 import { SuiObjectResponse } from "@mysten/sui/client";
 import logger from "./logger";
 
 /**
  * Redirects to the portal URL.
  */
 export function redirectToPortalURLResponse(
-    scope: URL, path: DomainDetails, portalDomainNameLength?: number
+    scope: URL,
```
nikos-kitmeridis, line 14: `scope` is not used, let's remove it.

portal/common/lib/url_fetcher.ts
```diff
@@ -176,12 +177,22 @@ export class UrlFetcher {
 
         logger.info("Successfully fetched resource!", { fetchedResourceResult: JSON.stringify(result) });
 
+        const quilt_patch_internal_id = result.headers.get("x-wal-quilt-patch-internal-id")
+		let aggregator_endpoint: URL;
```
nikos-kitmeridis, line 181: formatting looks a bit off in some editors due to the usage of tabs.
Let's stick with using spaces instead.

portal/common/lib/aggregator.ts
Tzal3x: Fair point! Updating it.

portal/common/lib/aggregator.ts
Tzal3x: Alright I reviewed and added these tests with a modification: `it('throws if aggregatorUrl ends with a slash', () => {` were no longer valid since we no longer throw an error but we just remove the end slash directly.

PR #537, opened 2025-08-28
feat(site-builder): Group files to quilts

nikos-terzo:
Introduces a new command `publish-quilts` under the feature "quilts" which groups the site resources into Quilts before uploading to Walrus.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -125,7 +125,7 @@ jobs:
         with:
           save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
       - name: Run tests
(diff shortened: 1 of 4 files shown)


site-builder/src/lib.rs
```diff
@@ -121,6 +121,21 @@ async fn run_internal(
                 .run()
                 .await?
         }
+        Commands::QuiltsPublish {
```
Tzal3x, line 124: Can we rename the `quilts-publish` to `publish-quilts` so that we follow the same convention as in walrus? 

In walrus they do `walrus store` or `walrus store-quilt`. The intuitive thing to do with sites would be `site-builder publish` and `site-builder publish-quilts`.

PR #542, opened 2025-09-04
fix(site-builder): add more checks for tx failure

wbbradley:


Tzal3x (approved):
Awesome, thank you @wbbradley !

PR #549, opened 2025-09-09
fix(sonarqube): integrate Javascript/Typescript tests with Sonarqube

Tzal3x:
1. For sonarqube to work we have to define the test directories. Sonarqube works better when source code and tests are in separate directories.
2. For JS/TS a [manual coverage exporter must be set up](https://docs.sonarsource.com/sonarqube-server/9.7/analyzing-source-code/test-coverage/javascript-typescript-test-coverage/). This means using [vitest's v8 coverage reporter](https://vitest.dev/guide/coverage.html#coverage-setup) along with the CI sonarqube tool. 
3. Updated the dependencies of the portal. This was especially needed for vitest because there was a version mismatch between it and the vitest-coverage-v8 resulting in errors while creating the test coverage report.


Tzal3x:
> I strongly recommend the following changes before merging 

@nikos-kitmeridis thank you for pointing this out. I made the change! ✅

PR #554, opened 2025-09-10
chore: remove publish_landing_page.sh script

Tzal3x:
This was a script we were using in the past
to publish the old landing page code.

Now the wal.app landing page code is in a
different directory, following a completely
different build/deploy process, making this
script useless.


nikos-terzo (approved):
💯🚮

PR #557, opened 2025-09-12
docs(site-builder): Improve the print_summary docs 

ioannischtz:
Improve the print_summary docs for testnet deployments.


site-builder/src/publish.rs
```diff
@@ -443,10 +443,15 @@ fn print_summary(
...
+            r#"⚠️ wal.app only supports sites deployed on mainnet.
+     To browse your testnet site, you need to self-host a portal:
+     1. For local development: http://{base36_id}.localhost:3000
+     2. For public sharing: http://{base36_id}.yourdomain.com:3000
 
-    (more info: https://docs.wal.app/walrus-sites/portal.html#running-the-portal-locally)"#,
+     📖 Setup instructions: https://docs.wal.app/walrus-sites/portal.html#running-the-portal-locally
+
+     💡 Tip: You may also bring your own domain (https://docs.wal.app/walrus-sites/bring-your-own-domain.html)
+            or find third-party hosted testnet portals."#,
```
ioannischtz, line 454: Not sure if we want to mention third-party hosted testnet portals here. 
Maybe if we had a list, that we could share?

PR #559, opened 2025-09-16
chore(site-builder): sew 387 create integration test for list directory

Tzal3x:
- [x] `site-builder --context=testnet publish --list-directory examples/snake --epochs=1` should publish a browsable site, but not the ignored fields defined in the ws-resources.json. The ignored resource paths should not be included in the index.html, and the resources should not be created on Sui/Walrus (i.e. they should not be browsable).
- [x] `site-builder --context=testnet list-directory examples/snake` should only generate a the new directory, containing the new index.html files to browse the directory, but also ignore the fields defined in ws-resources.json.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,3 @@ env/
 temp-landing-page/
 .env*.local
 **/coverage
-#examples/snake


site-builder/tests/list_directory.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) Mysten Labs, Inc.
```
nikos-kitmeridis, line 1: Why are the tests marked as ignored?

nikos-kitmeridis (approved):

site-builder/tests/list_directory.rs
Tzal3x: That's only for local development. Because whenever we make a commit, we want only the unit tests to run, not the integration tests. On the other hand we have a github action that runs the integration checks, so we have the benefits of both worlds without sacrificing our time during local dev.

PR #560, opened 2025-09-16
fix(common-lib): nullish coalescing for range header building

Tzal3x:
Fix the optionalRangeToHeaders( start: 0, end: 10 }) so that instead of returning bytes=-10 it should
return bytes=0-10.

This way it aligns with how range headers should be constructed. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range

Since I am already adding tests for the type/index.ts I included more tests to improve our coverage.

### DIFF
--- a/portal/common/lib/tests/types.test.ts
+++ b/portal/common/lib/tests/types.test.ts
@@ -1,0 +1,29 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { optionalRangeToHeaders, Range } from "@lib/types/index";
+import { describe, it, expect } from "vitest";
+
+describe("happy paths of optionalRangeToHeaders should", () => {
+    it.each([
+        [{ start: 0, end: 10 }, "bytes=0-10"],
+        [{ start: null, end: 10 }, "bytes=-10"],
+        [{ start: 10, end: null }, "bytes=10-"],
+        [null, undefined],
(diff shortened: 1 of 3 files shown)


portal/common/lib/tests/types.test.ts
```diff
@@ -0,0 +1,157 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+    isResource,
+    isVersionedResource,
+    optionalRangeToHeaders,
+    Range,
+    isRoutes,
+} from "@lib/types/index";
```
nikos-kitmeridis, line 10: can be simplified to `"@lib/types"`

portal/common/lib/tests/types.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    });
+
+    it.each([
+        [{ other_property: "value" }, "routes_list is missing"],
+        [{ routes_list: {} }, "routes_list is not a Map but an object"],
+        [{ routes_list: [] }, "routes_list is not a Map but an array"],
+        [{ routes_list: "not a map" }, "routes_list is not a Map but a string"],
+        [null, "obj is null"],
+        [undefined, "obj is undefined"],
+    ])("should return false when %s", (input, description) => {
```
nikos-kitmeridis, line 154: `description` is not used

nikos-kitmeridis (approved):

PR #564, opened 2025-09-23
fix(site-builder): Dynamically split move-commands into ptbs

nikos-terzo:
We are currently hardcoding the number of resources to include per PTB. This is erroneous, as having different number of information per resource results in a different number of move-calls per resource.


site-builder/src/site/manager.rs
```diff
@@ -402,19 +406,19 @@ impl SiteManager {
...
             let mut ptb = ptb.with_call_arg(&call_arg)?;
-            ptb.add_resource_operations(&updates.resource_ops[start..end])?;
+
+            ptb.add_resource_operations(&mut resources_iter)
+                .ok_if_limit_reached()?;
+            ptb.add_route_operations(&mut routes_iter)
+                .ok_if_limit_reached()?;
 
             let resource_result = self
                 .sign_and_send_ptb(ptb.finish(), self.gas_coin_ref().await?, &retry_client)
```
ioannischtz, line 424: we are asserting on line 383:  `assert!(built_ptb.commands.len() <= PTB_MAX_MOVE_CALLS as usize);`

 but not here inside the loop. could that be an issue?

ioannischtz: never mind, I guess the `ok_if_limit_reached()?` covers us.

site-builder/src/unit_tests/util.tests.rs
```diff
@@ -68,11 +68,11 @@ fn test_is_pattern_match() {
             path: "/test/foo-bar_baz.extension",
             expected: true,
         },
-        PatternMatchTestCase {
-            pattern: "[invalid",
-            path: "/file",
-            expected: false,
-        },
+        // PatternMatchTestCase {
```
ioannischtz, line 71: nit: leftover commented out code

site-builder/tests/publish.rs
```diff
@@ -60,3 +60,5 @@ async fn publish_snake() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+// TODO: Integration tests with a lot of routes.
```
ioannischtz, line 64: let's add this on separate PR

ioannischtz (approved):
Great work!
left some comments, but looking good.

site-builder/src/unit_tests/util.tests.rs
nikos-terzo: This is an erroring test from previous PR. It was just that our CI wasn't catching it because it was not properly set as test. Created a ticket for this.

PR #565, opened 2025-09-29
fix(portal): Typo, not passing args in getPropertiesWithContext

nikos-terzo:
Typo in setting up error predicate in logs resulted in not printing additional context given by the developer.


ioannischtz (approved):
Good catch

PR #568, opened 2025-09-30
test(site-builder): add checks for the frost cost estimation

ioannischtz:
test(site-builder): add checks for the frost cost estimation of the dry-run. 
Note1: needed the gag dev-dependency for this.
Note2: Had to put both test cases into one, due to gag BufferRedirect not working properly for parallel tests. Didn't want to add serial_test crate just for this.


nikos-terzo (approved):
Thank you @ioannischtz . We should sometime do an effort to add often used functions directly to `TestCluster` eg. `wal_coin_type()`, `active_address()`

PR #704, opened 2026-04-17
fix(site-builder): use active env, not envs.first(), for sitemap RPC

nikos-terzo:
SuiClientConfig::get_env(&None) returns envs.first(), not the active env. In sitemap this caused the Sui client to hit whichever env happened to be listed first in the user's wallet config instead of the one selected via --context / --wallet-env, surfacing as a misleading "Walrus package not found" error from the SDK.


ioannischtz (approved):
nice ✅

PR #705, opened 2026-04-21
fix(portal): count every request in ws_num_requests_made_counter

nikos-terzo:
Move the counter bump to the top of resolveDomainAndFetchUrl so it runs before SuiNS resolution. Previously, requests that failed resolution (and returned fullNodeFail) bypassed the counter, making the full_node_fail / requests_made ratio alert measure two different populations. Also drops the unused _requestId argument.


ioannischtz (approved):
✅

PR #709, opened 2026-04-28
fix: portal suins fn failures

nikos-terzo:
## Summary

The portal's FN-failure metric was getting bumped every time someone visited a URL with an unregistered SuiNS name. The `@mysten/sui` SDK throws an `ObjectError` with `code: "notExists"` when the FN authoritatively answers "this dynamic field does not exist" — that's a clean negative response, not an FN failure, but the `RPCSelector` was treating it as a transient error and retrying across all configured RPCs before bubbling up.

This PR teaches `RPCSelector` to recognize `notExists` and short-circuit:

- `rpc_selector.ts`: add `isNotExistsError` duck-type check (the SDK doesn't export `ObjectError`). In `invokeWithFailover`, return `status: "stop"` on `notExists` so we don't retry across RPCs. In `getNameRecord`, unwrap the resulting `AggregateError` and return `null` — the documented "name not registered" signal — instead of throwing.
- `url_fetcher.ts`: include the actual error in the log line when SuiNS resolution fails, so the next debugging pass isn't blind.

Tests cover both the RPC-selector path (`rpc_selector_get_name_record.test.ts`), the URL-fetcher path (`url_fetcher_resolve_object_id.test.ts`), and pin the SDK's exception shape against testnet (`suins_exception_shape.testnet.test.ts`) so we get a loud failure if `@mysten/sui` changes the error type.

## Test plan

- [x] `bun -F common test` passes
- [ ] Verify on a deployed portal that hitting an unregistered SuiNS name no longer increments the FN-failure counter
- [ ] Verify a real FN outage still increments the counter

### DIFF
--- a/portal/common/lib/tests/rpc_selector_get_name_record.test.ts
+++ b/portal/common/lib/tests/rpc_selector_get_name_record.test.ts
@@ -1,0 +1,136 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { RPCSelector } from "@lib/rpc_selector";
+import { PriorityExecutor, PriorityUrl } from "@lib/priority_executor";
+
+/**
+ * Tests that RPCSelector.getNameRecord handles ObjectError(notExists)
+ * correctly: returns null (no retries) instead of propagating the error.
+ *
+ * The @mysten/sui SDK throws ObjectError(code: "notExists") when the FN
+ * cleanly responds that a dynamic field doesn't exist — which is exactly
+ * what happens for any unregistered SuiNS name. This is an authoritative
+ * answer (name registration is determinate on-chain state), so the
+ * RPCSelector should stop immediately and return null.
+ */
+
+/** Mirrors @mysten/sui ObjectError shape (not publicly exported). */
+class ObjectErrorLike extends Error {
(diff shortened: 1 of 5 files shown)


portal/common/lib/tests/rpc_selector_get_name_record.test.ts
```diff
@@ -0,0 +1,136 @@
...
+            return result.value;
+        });
+
+        // Build RPCSelector and replace its executor with our spy.
+        const rpcSelector = new RPCSelector(urls, "testnet");
+        (rpcSelector as unknown as { executor: PriorityExecutor }).executor = executor;
+
+        // The getNameRecord catch block should detect notExists and return null.
+        // But we need invokeWithFailover to actually run... Let's take a
+        // different approach: test the full flow by mocking fetch.
```
ioannischtz: > ...Let's take a different approach

THis reads like the parts you set up above (callCount, invokeSpy) are not actually used? but we are going with "the different approach" of mocking the fetch?

let's cleanup this test file

ioannischtz (approved):
ty @nikos-terzo , implementation LGTM

left one comment for the test

portal/common/lib/tests/rpc_selector_get_name_record.test.ts
nikos-terzo: Thanks @ioannischtz ! Resolving...

PR #713, opened 2026-05-19
chore(site-builder): Update sui walrus deps

nikos-terzo:
Walrus is pinned to commit f39a84f0 because it includes walrus#3389
(owner/version/digest on ObjectChangeEntry), which we need before a
tagged Walrus release ships with it.

Adapts site-builder to the new walrus-sui execute API:
- sign_and_send_ptb returns ExecuteTransactionResponse and passes
  Duration::ZERO for the new checkpoint_wait_timeout arg
- update_cache_from_~~effects~~obj_changes now walks Vec<ObjectChangeEntry>
- get_site_id_from_response reads effects_status + object_changes
- update_site returns Option<...> (no Default on the new response)
- dry_run_ptb builds its own SuiClient since RetriableSuiClient no
  longer exposes the inner client
- tokio bumped to =1.52.1 to match walrus's pin

### DIFF
--- a/site-builder/src/site/manager.rs
+++ b/site-builder/src/site/manager.rs
@@ -444,9 +444,8 @@ impl SiteManager {
             gas_price, // Use actual reference gas price
         );
 
-        // TODO(grpc-migration): Replace with gRPC-backed simulate/dev_inspect once the
-        // walrus-sui crate adds support. RetriableSuiClient no longer exposes the inner
-        // SuiClient, so we build a standalone JSON-RPC client from the wallet's RPC URL.
+        // TODO(grpc-migration): Migrate to gRPC-backed simulate/dev_inspect.
+        // Builds a standalone JSON-RPC client from the wallet's RPC URL.
         let rpc_url = &self
             .wallet
             .config
--- a/site-builder/src/site/manager.rs
+++ b/site-builder/src/site/manager.rs
@@ -11,7 +11,6 @@ use anyhow::{anyhow, bail, Context, Result};
 use sui_keys::keystore::AccountKeystore;
 use sui_sdk::{
     json_rpc_error::TRANSACTION_EXECUTION_CLIENT_ERROR_CODE,
-    rpc_types::SuiTransactionBlockResponse,
     wallet_context::WalletContext,
 };
 use sui_types::{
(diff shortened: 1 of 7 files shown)


ioannischtz (approved):
ty @nikos-terzo , looks good from my end ✅

site-builder/src/site/manager.rs
```diff
@@ -441,13 +445,17 @@ impl SiteManager {
         );
 
         // TODO(grpc-migration): Replace with gRPC-backed simulate/dev_inspect once the
-        // walrus-sui crate adds support. Currently goes through deprecated
-        // get_current_client() → sui_client() → read_api() → dev_inspect_transaction_block().
+        // walrus-sui crate adds support. RetriableSuiClient no longer exposes the inner
```
wbbradley:
```
        // walrus-sui crate adds support. RetriableSuiClient no longer exposes the inner
```
I think we can avoid putting past-related comments in. Future code readers will not understand why "RetriableSuiClient no longer exposes" is relevant to anything since it is an archeological footnote that can be found in git if future devs are interested.

site-builder/src/site/manager.rs
```diff
@@ -441,13 +445,17 @@ impl SiteManager {
...
+        // SuiClient, so we build a standalone JSON-RPC client from the wallet's RPC URL.
+        let rpc_url = &self
```
wbbradley, line 449: We should consider how we will migrate away from JSON-RPC. Although, in this case, I believe this is actually using gRPC under the hood. @bmwill is this correct?

bmwill: No this is still using jsonrpc, you can see below its making jsonrpc calls

wbbradley: Yes, I saw those calls, but i thought I recalled that there was some redirection going on at some layer in sui_sdk, maybe it's just the WalletContext specific methods. Thanks. In this case, we'll need to follow up on this PR to remove JSON-RPC from Walrus Sites further.

bmwill: yes its just WalletContext specific calls, this is constructing its own jsonrpc client

site-builder/src/site/manager.rs
nikos-terzo: Fixed, thanks @wbbradley !

site-builder/src/site/manager.rs
nikos-terzo: Thanks, I have created a respective follow-up ticket for this. Resolving convo for this PR.

PR #714, opened 2026-05-20
fix(portal): size Bun idleTimeout from aggregator retry budget

nikos-terzo:
When the aggregator chain takes longer than Bun.serve's default 10s
idleTimeout, Bun closes the inbound socket mid-flight and the upstream
proxy returns its own (generic) error body to the user instead of the
portal's aggregatorFail() page.

## Changes

- Apply a 10s per-attempt timeout to the outbound aggregator fetch via
  `AbortSignal.timeout`, so each attempt has a known cap.
- Add `PriorityExecutor.worstCaseDurationMs(perAttemptMs)` to compute
  the worst-case duration from the configured URLs/retries.
- Size `Bun.serve` `idleTimeout` from that, plus a small headroom,
  clamped to a sensible max (fits under typical upstream proxy/CDN
  budgets).
- Both the per-attempt timeout and the idleTimeout cap are configurable
  via env vars `AGGREGATOR_REQUEST_TIMEOUT_MS` and
  `PORTAL_IDLE_TIMEOUT_MAX_S`.
- Sanitize-and-log resolved portal config at startup (secret fields
  shown as `[set]`/`[unset]`).

## ⚠️ Behavioral change

Previously, a slow-but-OK aggregator that took 30s to return 200 would
still succeed. With this change, any single attempt is aborted at 10s
(default) and the request fails over to the next aggregator URL. The
chain still completes successfully against fast aggregators; the only
requests that change are those waiting on a slow upstream.

## Tuning for your deployment

The defaults are reasonable for a typical setup. If yours is different,
edit the env vars:

- `AGGREGATOR_REQUEST_TIMEOUT_MS` (default `10000`) — per-attempt
  aggregator fetch timeout. Raise it if your aggregators legitimately
  need longer than 10s; lower it to fail over sooner.
- `PORTAL_IDLE_TIMEOUT_MAX_S` (default `100`) — upper bound on the
  Bun server's `idleTimeout`. Set it below whatever upstream proxy or
  CDN sits in front of the portal.

The portal logs the resolved `idleTimeoutS` at startup so you can see
exactly what it picked given your config.

## Tests

- Unit tests for `worstCaseDurationMs` and `worstCaseAggregatorChainMs`
  in the common-lib vitest suite.
- New e2e workflow `e2e-aggregator-timeout.yml` that stands up
  mock-aggregator + portal + envoy, asserts the portal returns its own
  503 `aggregatorFail()` response under the fixed timing budget.


scripts/e2e-aggregator-timeout/e2e.sh
```diff
@@ -0,0 +1,95 @@
...
+mkdir -p "$LOG"
+rm -f "$LOG"/*.log "$LOG"/response.body
+
+cleanup() {
+    lsof -ti :3000 -ti :4000 -ti :8080 2>/dev/null | xargs -r kill -9 2>/dev/null || true
+}
+trap cleanup EXIT
+cleanup
+
+command -v envoy >/dev/null || { echo "envoy not in PATH"; exit 1; }
```
nikos-terzo: Note to reviewer.

envoy is not in scope for this repo, but we use it to reproduce the exact message that the users were receiving on production. Should we remove it from this repro?

portal/server/index.ts
```diff
@@ -8,13 +8,55 @@ import { genericError } from "@lib/http/http_error_responses";
...
+ */
+const DEFAULT_IDLE_TIMEOUT_MAX_S = 100;
+const idleTimeoutMaxS = Number(process.env.PORTAL_IDLE_TIMEOUT_MAX_S) || DEFAULT_IDLE_TIMEOUT_MAX_S;
+
+// Bun.serve closes the inbound socket after `idleTimeout` seconds. If that
+// fires before the portal has returned a response, an upstream proxy may
+// substitute its own (less helpful) error body. Size it from the aggregator
+// retry budget so the chain always has room to complete and return our own
+// aggregatorFail() response.
+const idleTimeoutS = Math.min(
```
ioannischtz, line 53: idleTimeout is sized from the aggregator chain only. A request also runs several RPC failover chains (resolveObjectId, getRoutesAndRedirects, one fetchResource per fetchUrl), each rpcRequestTimeoutMs×retries (~15s default).|

Is that something we need to take into account as well? Should we add `executor.worstCaseDurationMs(rpcRequestTimeoutMs) into the budget` ?

portal/server/index.ts
```diff
@@ -8,13 +8,55 @@ import { genericError } from "@lib/http/http_error_responses";
...
+// hashing, and small per-attempt variance.
+const IDLE_TIMEOUT_HEADROOM_S = 10;
+/**
+ * Default upper bound (seconds) on the computed idleTimeout. Overridable via
+ * the `PORTAL_IDLE_TIMEOUT_MAX_S` env var. Set this below any upstream proxy
+ * or CDN request timeout sitting in front of the portal — depending on the
+ * dependent service's timeout, it may need to be lower than 100s (e.g.
+ * Cloudflare's default free-tier proxy timeout is 100s).
+ *
+ * Bun's `idleTimeout` accepts at most 255s; values above that are rejected
```
ioannischtz: let's add Bun's ceiling in the calculation.

ioannischtz: Maybe `Math.min(…,255)` + a `logger.warn` when the budget exceeds the cap.

portal/common/lib/src/url_fetcher.ts
```diff
@@ -340,7 +365,15 @@ export class UrlFetcher {
         const start = Date.now();
 
         try {
-            const response = await fetch(url, { headers });
+            // NOTE: As of early 2026, Bun's AbortSignal rejects the fetch
+            // promise on timeout but the underlying request still completes in
+            // the background (same caveat as in rpc_selector.ts). Under load,
+            // repeated timeouts can hold sockets open until the upstream
+            // naturally responds. This may improve in future Bun versions.
+            const response = await fetch(url, {
```
ioannischtz, line 381: Only the per-attempt timeout is wired; the inbound request.signal isn't. On client/proxy disconnect the portal still runs the whole retry chain for a response nobody reads.

Maybe something like `AbortSignal.any([request.signal, AbortSignal.timeout(aggregatorTimeoutMs)])`

scripts/e2e-aggregator-timeout/e2e.sh
```diff
@@ -0,0 +1,95 @@
...
+[[ -d "$PORTAL_DIR/node_modules" ]] || (cd "$PORTAL_DIR" && bun install --frozen-lockfile)
+
+echo "starting mock-aggregator (MODE=${MODE:-fail503})..."
```
ioannischtz: From Claude:

fail503 returns 503 at 8s, inside the 10s abort window → the fetch resolves and is classified retry-same; the AbortSignal.timeout→catch→retry-next branch (url_fetcher.ts:439) never runs. The headline fix is unguarded — deleting signal:
would still pass CI. sleep/headers_then_hang modes are dead. Add an e2e leg with MODE=sleep SLEEP_MS=30000 asserting the portal returns its own 503.

portal/common/lib/tests/url_fetcher_error_handling.test.ts
```diff
@@ -258,3 +258,35 @@ describe("aggregator error handling with priority executor", () => {
...
+    }
+
+    it("returns the executor's worst-case chain at the configured per-attempt timeout", () => {
+        const items: PriorityUrl[] = [
+            { url: "https://a.com", retries: 2, metric: 100 },
+            { url: "https://b.com", retries: 0, metric: 200 },
+        ];
+        const fetcher = makeFetcher(items);
+
+        // Library applies a 10s default per-attempt timeout (DEFAULT_AGGREGATOR_TIMEOUT_MS).
```
ioannischtz: From Claude:

Non-hermetic: holds only while aggregatorTimeoutMs===10000, but vitest loadEnv(mode,cwd,'') injects all env vars, so a shell exporting AGGREGATOR_REQUEST_TIMEOUT_MS flips it. Also mirrors the impl literal. import { aggregatorTimeoutMs} and derive the expectation.

.github/workflows/e2e-aggregator-timeout.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  e2e:
+    name: E2E aggregator timeout
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - uses: oven-sh/setup-bun@735343b667d3e6f658f44d0eca948eb6282f2b76 # v2.0.2
```
ioannischtz, line 40: No bun-version → installs latest, but prod is oven/bun:1.2.5 and this test depends on Bun's version-specific AbortSignal/idleTimeout semantics (per the code's own NOTE)

Should pin to prod version, but this is also good opportunity to discuss/schedule a Bun update for prod?

portal/common/lib/src/url_fetcher.ts
```diff
@@ -35,6 +35,22 @@ type AggregatorResult =
...
+ * aggregator response longer than this is aborted and the request fails over
+ * to the next aggregator URL.
+ *
+ * TODO: any new env-driven knob in common/lib must also be added to the
+ * worker's webpack DefinePlugin (`portal/worker/webpack.config.common.js`),
+ * or the worker bundle will read a bare `process.env.X` that is undefined
+ * at runtime. A central registry would prevent that drift.
+ */
+const DEFAULT_AGGREGATOR_TIMEOUT_MS = 10_000;
+export const aggregatorTimeoutMs =
```
ioannischtz: From Claude: 

0/NaN/negatives all silently fall to DEFAULT; a negative reaches AbortSignal.timeout(-1) unvalidated. Inconsistent with the loader's throw-on-invalid. Validate >0 + warn.

portal/server/.env.mainnet.example
```diff
@@ -5,17 +5,34 @@
 # Env vars below are supported as overrides — if both YAML and env var
 # are set for a field, the env var wins.
```
ioannischtz, line 7: PROMETHEUS_EXPORTER_PORT is active in the testnet example but missing here — add for parity

ioannischtz (commented):

portal/server/index.ts
nikos-terzo: An already configurable limit is set (`DEFAULT_IDLE_TIMEOUT_MAX_S`). It is just that the configurable limit makes sense only if lower than bun's one. Maybe we can just warn the user if they set something larger than 255 seconds.

portal/server/index.ts
nikos-terzo: Correct. The thinking was that sui rpc-requests usually respond a lot faster that's why we are not taking them into account. I do not want the sui-rpc requests increasing this timeout to even larger values. I will investigate this further.

portal/common/lib/src/url_fetcher.ts
nikos-terzo: Actually this requires threading the signal from the initial request all the way to aggregator fetch.
This is indeed something nice to have, but will introduce larger changes, so I say we handle this in a follow-up.
Request's abort signal is something we were ignoring previous to this PR anyway.

portal/server/index.ts
nikos-terzo: The rpc chain with 7 sec timeout would greatly influence the idle timeout. We do not want that, as empirically, it rarely timeouts. Addressed with a code-comment at f793c77

portal/server/index.ts
nikos-terzo: Addressed at 6ea42d7

scripts/e2e-aggregator-timeout/e2e.sh
nikos-terzo: Addressed at a157045 (refactor), 8fc2ff7 (new timeout test)

portal/common/lib/tests/url_fetcher_error_handling.test.ts
nikos-terzo: Addressed at 2859220

.github/workflows/e2e-aggregator-timeout.yml
nikos-terzo: Addressed at 8ea17a4

portal/common/lib/src/url_fetcher.ts
nikos-terzo: Addressed at ba16578

portal/server/.env.mainnet.example
nikos-terzo: Addressed at 23ceefd

ioannischtz (approved):
ty @nikos-terzo 

Other than the CI failures, lgtm

PR #720, opened 2026-06-29
feat(portal): glob route/redirect matcher behind a flag, with a pattern cap

nikos-terzo:
Replaces the portal's regex route matcher with a small purpose-built glob matcher, behind a feature flag (default off), plus a pattern-complexity cap that keeps matching cost predictable.
Patterns come from on-chain site objects, so match cost should stay bounded however they're written.

No picomatch:
- It compiles patterns to a regex whose worst-case cost is hard to bound.
- The hand-rolled matchGlob guarantees ≤1 `*` per segment and ≤1 `**` per pattern, so matching is linear in the path length by construction.
- picomatch is removed.

**Grammar**: `*` matches within a segment, `**` across segments, everything else is literal. Legacy routes are rewritten to keep their reach:
- bare `*` → `/**` 
- trailing `/*` → `/**/*`
- mid `*` stays single-segment
- existing `**` untouched

**Pattern cap (always on, not flagged)**:
- runs before every match, for routes + redirects
- over-cap patterns are skipped and logged, never matched

**Behavior (flag on)**: most-specific match wins (most literals, then fewest wildcards) instead of longest-wins:
- an exact `/dashboard/` can now beat `/dashboard/*`
- a mid `*` like `/forms/*/admin` no longer crosses `/`

**Redirects (always, not flagged)**: move to the same glob matcher + most-specific tiebreaker, regardless of the flag

**Flag**: `ENABLE_GLOB_ROUTING` / `enable_glob_routing` / `enableGlobRouting`, default off, wired through server + worker.

**Test**: `cd portal && bun -F common test` — new _path_patterns.test.ts_, updated _routing.test.ts_.


portal/common/lib/src/path_patterns.ts
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Validating and matching owner-supplied route/redirect patterns.
+ *
+ * Grammar: `*` matches within one path segment, `**` (a whole segment) matches
+ * across segments, and every other character is a literal. `*` is reserved, so a
+ * literal `*` in a path can't be targeted; every other URL path character
+ * (`( ) [ ] ! +` ...) just works.
+ *
+ * Wildcards are capped so a crafted pattern can't freeze the event loop (ReDoS):
```
ioannischtz: **ReDoS note holds for the glob path only**

Flag-off, `MAX_STARS=2` still admits `**foo` → `^.*.*foo$` (quadratic on long paths).
  Suggest rewording: the cap **bounds the regex branch to quadratic**, it doesn't eliminate the stall.

ioannischtz: (Claude)
**`main.ts` — add a request path-length cap**

  No in-app bound on `path`, so the flag-off quadratic above is reachable with long paths

(Or is this covered in the smart contract?)

portal/worker/webpack.config.common.js
ioannischtz, line 26: **worker YAML drops `enable_glob_routing`**

  `loadYamlDefaults()` maps every other key but this one, so a YAML-only
  `enable_glob_routing: true` resolves to `undefined` → worker silently stays **off**
  (the worker `.env` example points operators at this YAML). Fix:

  ```js
  if (yaml.enable_glob_routing !== undefined)
      defaults.ENABLE_GLOB_ROUTING = String(yaml.enable_glob_routing);

portal/common/lib/src/url_fetcher.ts
```diff
@@ -171,18 +171,7 @@ export class UrlFetcher {
...
-            );
-            if (regexOnlyPatterns.length > 0) {
-                logger.warn("Route patterns match via regex but not via glob (picomatch)", {
-                    subdomain: parsedUrl.subdomain,
-                    siteObjectId: resolvedObjectId,
-                    path: parsedUrl.path,
-                    regexOnlyPatterns,
-                });
-            }
+            const matchingRoute = this.wsRouter.matchPathToRoute(parsedUrl.path, routes);
```
ioannischtz, line 192: (Claude):

 **`url_fetcher.ts` — removed the only in-prod glob divergence signal**

  The dropped `regexOnlyPatterns` warning was the canary for sites that match under regex
  but not glob — i.e. what breaks when `ENABLE_GLOB_ROUTING` flips. Without it, only offline
  corpus replay catches drift. Keep a shadow-compare during the migration window, or state
  explicitly that corpus replay is the sole net.

portal/common/lib/src/path_patterns.ts
```diff
@@ -0,0 +1,190 @@
...
+ * on. Under the regex a `*` became `.*` and crossed `/`, so:
+ *  - a bare `*` matched everything, and becomes a root globstar `/**`;
+ *  - a trailing `/` then `*` matched paths one or more levels below the prefix;
+ *    the regex needed that slash, so it never matched the bare prefix. It
+ *    becomes a globstar plus a required segment, keeping that reach without
+ *    shadowing an exact route for the prefix itself.
+ * A `*` in the middle of a pattern stays within its segment, and a pattern that
+ * already uses `**` is returned unchanged.
+ */
+export function regexToGlobPattern(pattern: string): string {
```
ioannischtz, line 146: **`regexToGlobPattern` only rewrites slash-glued `/*`**

  A glued `/api*` falls through unchanged → glob matches single-segment, but the old regex
  `^/api.*$` crossed `/`. Silent narrowing for any site authoring `/api*`-style routes.
  Add to the on-chain corpus replay before flipping the flag.

portal/common/lib/src/path_patterns.ts
```diff
@@ -0,0 +1,190 @@
...
+ * (`( ) [ ] ! +` ...) just works.
+ *
+ * Wildcards are capped so a crafted pattern can't freeze the event loop (ReDoS):
+ *  - glob (redirects always; routes when the flag is on): each segment has at
+ *    most one `*` or is a whole-segment `**`, and at most one `**` per pattern.
+ *    Those two guarantees let `matchGlob` anchor without any backtracking.
+ *  - legacy regex (routes, flag off): the pattern is passed to `RegExp` with `*`
+ *    turned into `.*` (crossing `/`), so it accepts only plain path characters
+ *    plus `*` (the wildcard) and `.` (today's any-char). Every other regex
+ *    metacharacter is rejected — they crash `RegExp` (`( ) [ ] { }`), backtrack
```
ioannischtz: **nit: which chars crash `RegExp`**

  Only `(`, `)`, `[` actually throw; `]`, `}`, `{` are accepted as literals. Trim the parenthetical.

portal/common/lib/src/routing.ts
```diff
@@ -119,16 +135,43 @@ export class WalrusSitesRouter {
...
+    private matchGlobEntry<V>(
+        entries: Iterable<[string, V]>,
+        path: string,
+        kind: "route" | "redirect",
+    ): V | undefined {
+        let winner: { glob: string; value: V } | undefined;
+        for (const [glob, value] of entries) {
+            const validation = validateGlobPattern(glob);
+            if (!validation.ok) {
+                logger.warn(`Skipping unsafe ${kind} pattern`, {
```
ioannischtz: (Claude):  

**`routing.ts:162` — log skipped patterns once per site, not per request**

  Invalid patterns are static per-site, but this warns once **per pattern, per request**, each
  carrying the full path. Log once per site with a count. (Not a security issue — hygiene.)

portal/common/lib/tests/routing.test.ts
```diff
@@ -143,7 +143,7 @@ const testCases = [
```
ioannischtz, line 137: **`tests/routing.test.ts` — cover the behavior _changes_, not just the new engine**

  The suite proves flag-on correctness but never asserts the flag-off→flag-on deltas. Add:

  - **exact beats wildcard** (bug-fix 1): routes `{ "/foo/*": A, "/foo/": B }`, path `/foo/`
    — flag-off → `A` (longest-wins), flag-on → `B` (most-specific).
  - **mid-`*` narrows** (bug-fix 3): `/forms/*/admin` vs `/forms/a/b/admin`
    — flag-off matches (regex crosses `/`), flag-on `undefined`.
  - **glued trailing star**: `/api*` vs `/api/x/y` — flag-off matches, flag-on `undefined`.
  - **redirect extended-glob→literal**: `/x/{a,b}` no longer matches `/x/a`.
  - **skip-warn count**: N invalid patterns → N `logger.warn` per request (spy).

ioannischtz (commented):

portal/worker/webpack.config.common.js
nikos-terzo: Fixed in 5f2cf838dae18453c9b2d9fe8fcae8d6d62857d4 — `loadYamlDefaults()` now maps `enable_glob_routing` (`!== undefined` + `String()`, same shape as the server loader; the worker still decides truthiness with `=== "true"` at the parse point).

The commit also adds comments to the `.env`/portal-config example files documenting boolean parsing: env vars enable only on the exact string `"true"`, YAML needs an unquoted boolean.

Opened SEW-1043 (low prio) to unify boolean flag parsing across server/worker env+YAML with a shared helper — today the three paths each behave slightly differently.

portal/common/lib/src/url_fetcher.ts
nikos-terzo: Re-introduced in 3c056c8af06c1a5052c4e7c5d09dc61f61f5f651. With the flag off, `matchPathToRoute` now also runs the glob matcher and warns ("Route target will change when glob routing is enabled", with path + both targets) when the two targets differ.

Two properties worth noting:

- The glob side goes through `regexToGlobPattern` first, so paths the compatibility rewrite already keeps working stay silent — it only warns on real post-flip divergence.
- It compares targets, not patterns, so two different patterns resolving to the same target don't warn either.

The flag-on branch and the canary share one helper (`matchRouteViaGlob`), so there's a single glob code path; both get deleted together with the flag in SEW-1001. Covered by two new tests (glued `/api*` diverges and warns; an agreeing match stays silent).

portal/common/lib/src/path_patterns.ts
nikos-terzo: Intended behavior: under globs, `*` staying within one segment is the sensible semantics for `/foo*` — only `**` crosses `/`. The rewrite deliberately widens just the two spellings sites used as catch-alls under the old regex (bare `*` and trailing `/*`); a glued `/foo*` reads as "files in this directory starting with foo", and the old cross-slash reach was an artifact of `*` → `.*`.

Whether any real site depends on that artifact is exactly what the two nets are for: the shadow-compare canary (3c056c8) will log it from production traffic while the flag is off, and the SEW-1001 corpus replay covers it offline. If one turns up we can widen the rewrite then.

portal/common/lib/src/path_patterns.ts
nikos-terzo: On the path-length cap — measured on Bun (JSC), no in-app cap needed:

- Bun's HTTP parser hard-caps the request head at 16KiB: max accepted path ≈ 16,2xx chars, 431 beyond, our code never runs. Percent-decoding only shrinks a path.
- At that cap, the worst pattern the validator admits (2 stars, e.g. `^.*.*foo$`) is a clean ~120ms `false` — bounded quadratic, no throw, no stack blow-up. Exponential blow-up needs the grouping/quantifier metacharacters the validator rejects.

Not covered by the contract (patterns are on-chain, the path is the visitor's request) — the runtime cap is the bound, and it's ~120ms worst case for the flag-off window only.

portal/common/lib/tests/routing.test.ts
nikos-terzo: Added in c2f2b790825c9ca3b78626905c722fb2795b624d — a "flag-off -> flag-on behavior deltas" describe with all five cases, each asserting both matchers in the same test:

- `/foo/` + `/foo/*` on `/foo/`: flag-off `/catch.html` (longest-wins), flag-on `/exact.html` (most-specific)
- `/forms/*/admin` vs `/forms/a/b/admin`: flag-off matches, flag-on `undefined`
- glued `/api*` vs `/api/x/y`: flag-off matches, flag-on `undefined`
- redirect `/x/{a,b}`: no longer matches `/x/a`, still matches itself literally
- 2 invalid patterns -> 2 warns per request (4 after a second request) — pins the current per-request behavior; will be updated with the log-once-per-site change

portal/common/lib/src/routing.ts
nikos-terzo: Fixed in 53861ceccd796cbd4f400b68fcf80b20841c9893. I didn't want to introduce the complexity of adding global state in the service (a warned-patterns set living for the whole portal lifetime), so I compromised with one INFO log per request: skipped patterns are collected during matching and logged as a single line with count + pattern + reason, and the request path is dropped from the payload so the lines dedupe downstream.

Applied in both matchers (the legacy regex one goes away with the flag in SEW-1001); test updated to assert the single aggregated line.

portal/common/lib/src/path_patterns.ts
nikos-terzo: Reworded in 583dbc906032362d22308c6b9aa194bb271b2615 — the header now says the regex branch is "capped at two `.*`s — worst case quadratic, a bounded stall rather than a freeze". We also simplified the whole comment while at it (20 lines -> 13); the per-matcher bullet list is gone.

portal/common/lib/src/path_patterns.ts
nikos-terzo: Resolved in 583dbc906032362d22308c6b9aa194bb271b2615 by simplifying the header comment — the metacharacter enumeration is dropped entirely, so there's no wrong list to trim anymore.

ioannischtz (approved):
ty @nikos-terzo , looking good ✅

PR #721, opened 2026-06-30
feat: portal migrate to GRPC

nikos-terzo:


portal/common/lib/src/rpc_selector.ts
```diff
@@ -111,88 +134,48 @@ export class RPCSelector implements RPCSelectorInterface {
...
             } catch (error) {
                 const wrappedError = error instanceof Error ? error : new Error(String(error));
-                // ObjectError(notExists) means the FN authoritatively answered
-                // "this object does not exist" — retrying won't change the
-                // answer.
-                if (isNotExistsError(error)) {
-                    return { status: "stop", error: wrappedError };
-                }
                 logger.warn("RPC call failed", { url, error: formatError(error) });
                 return { status: "retry-same", error: wrappedError };
```
ioannischtz, line 149: **The authoritative-stop path is gone: every thrown error is now `retry-same`.**

Fine today under gRPC — misses arrive as array elements, not throws. But a *deterministic* thrown error (e.g. `RpcError INVALID_ARGUMENT` from a malformed id) now sweeps the whole failover list before failing: ~6 attempts × 7s worst case with the example config.

No action needed while ids are derived/validated upstream — flagging as a latent latency cliff if a future caller ever passes user-influenced ids.

portal/common/lib/src/rpc_selector.ts
```diff
@@ -48,18 +47,42 @@ class WrappedSuiClient extends SuiJsonRpcClient {
...
+export function isObjectNotFoundError(error: unknown): boolean {
+    if (!(error instanceof Error)) {
+        return false;
+    }
+    // Transport errors (RpcError) always carry a gRPC status `code` string; a
+    // genuine missing-object error has none.
+    if (typeof (error as { code?: unknown }).code === "string") {
+        return false;
+    }
+    return /^Object 0x[0-9a-f]+ not found/i.test(error.message);
```
ioannischtz, line 65: **The drift guards pin false negatives (a real miss keeps being detected); nothing pins false positives — a transient error misread as an authoritative miss would 404 a live site instead of failing over.**

The `code`-is-string shield is the only defense on that side. Worth pinning with direct unit cases:

- `Error("Object 0xab not found")` **with** `code: "NOT_FOUND"` → `false` for both classifiers
- plain `Error("Not Found")` — no code, no id → `false`
- `Error("Object 0xzz not found")` — non-hex id → `false` (pins regex strictness)
- non-`Error` inputs — `"Object 0x1 not found"` string, `null` → `false`
- `Error("name.sui is not registered")` → `true` for `isNameNotRegisteredError`; same message plus a `code` → `false`
- `AggregateError` whose *message* contains `"Object 0x… not found"` → `false` (pins the `^` anchor)
- `"OBJECT 0X6 NOT FOUND"` → decide: `true` pins the `i` flag as intended, or drop the flag

portal/common/lib/src/rpc_selector.ts
```diff
@@ -17,18 +12,22 @@ const DEFAULT_RPC_REQUEST_TIMEOUT_MS = 7_000;
...
-    getDynamicFieldObject(input: GetDynamicFieldObjectParams): Promise<SuiObjectResponse>;
+    multiGetObjects<Include extends SuiClientTypes.ObjectInclude>(
+        objectIds: string[],
+        include: Include,
+    ): Promise<SuiClientTypes.GetObjectsResponse<Include>["objects"]>;
 }
 
-class WrappedSuiClient extends SuiJsonRpcClient {
+class WrappedSuiClient extends SuiGrpcClient {
     private url: string;
```
ioannischtz, line 27: Nit: `url`/`getURL()` duplicate the `baseUrl` the base `SuiGrpcClient` was constructed with — if the base client exposes it, these can go.

portal/common/lib/src/rpc_selector.ts
```diff
@@ -17,18 +12,22 @@ const DEFAULT_RPC_REQUEST_TIMEOUT_MS = 7_000;
...
-    multiGetObjects(input: MultiGetObjectsParams): Promise<SuiObjectResponse[]>;
-    getDynamicFieldObject(input: GetDynamicFieldObjectParams): Promise<SuiObjectResponse>;
+    multiGetObjects<Include extends SuiClientTypes.ObjectInclude>(
+        objectIds: string[],
+        include: Include,
+    ): Promise<SuiClientTypes.GetObjectsResponse<Include>["objects"]>;
 }
 
-class WrappedSuiClient extends SuiJsonRpcClient {
+class WrappedSuiClient extends SuiGrpcClient {
```
ioannischtz, line 26: potential tech-debt improvement: Migrate this to the client-extension system with `CoreClient.$extend`.

obviously out-of-scope for this PR

portal/common/lib/src/resource.ts
```diff
@@ -74,39 +73,50 @@ export class ResourceFetcher {
 
         seenResources.add(objectId);
 
-        const redirectId = checkRedirect(primaryObjectResponse);
-        if (redirectId) {
-            return this.fetchResource(redirectId, path, seenResources, depth + 1);
+        // Note: if the site object was deleted, its dynamic fields survive on
```
ioannischtz, line 76: **Reads as if the code is about to handle orphans — it actually explains why there is no early-404 branch.** Suggest stating the invariant directly:

```suggestion
        // Deliberate: a missing site object does NOT early-return 404. If the site
        // was deleted, its dynamic fields survive on chain as fetchable orphans, so
        // the fetch below may still succeed and serve the deleted site's content —
        // pre-existing behavior, intentionally preserved by the gRPC migration.
```

(or something like that)

portal/common/lib/src/resource.ts
```diff
@@ -74,39 +73,50 @@ export class ResourceFetcher {
...
+        // chain as orphans and stay fetchable by derived ID — so the resource
+        // fetch below can still succeed and serve a deleted site's leftovers.
+
+        // A miss (the site object wasn't found) carries no Display to redirect on.
+        if (!(primaryObjectResponse instanceof Error)) {
+            const redirectId = checkRedirect(primaryObjectResponse);
+            if (redirectId) {
+                return this.fetchResource(redirectId, path, seenResources, depth + 1);
+            }
+        }
```
ioannischtz, line 93: **When the primary object is an `Error`, the redirect check is skipped with no log** — a deleted/nonexistent site is indistinguishable in logs from a missing path (the later miss log reads as a path-level 404).

One line makes those incidents debuggable:

```ts
} else {
    logger.info("Site object not found", { objectId, reason: primaryObjectResponse.message });
}
```

portal/common/lib/src/redirects.ts
```diff
@@ -33,19 +33,21 @@ export function redirectToAggregatorUrlResponse(blobId: string, aggregatorUrl: s
 
 /**
  * Checks if the object has a redirect in its Display representation.
+ *
+ * IMPORTANT: this reads the *rendered* Display returned by gRPC
```
ioannischtz, line 37: **Verified on mainnet:** the docs.sui site renders identical Display fields over gRPC and JSON-RPC, so `Display<Site>` is in the v2 registry — and the [Display V2 announcement](https://blog.sui.io/display-v2-mainnet/) confirms a one-time system snapshot auto-migrated all pre-existing v1 Displays. Exposure is only Displays created via the v1 API *after* the snapshot. Suggest encoding that:

```suggestion
 * IMPORTANT: this reads the *rendered* Display returned by gRPC
 * (`core.getObjects`), which renders only Display **v2** (the `0xd` registry).
 * Sui's one-time system snapshot auto-migrated all pre-existing v1 Displays to
 * v2 (verified: mainnet `Display<Site>` renders over gRPC). A Display created
 * via the legacy v1 API *after* that snapshot is not in the registry: it
 * renders as `undefined` here and its redirect silently won't fire until
 * migrated (`display_registry::v1_to_v2`). JSON-RPC rendered v1 too — which is
 * why this was invisible pre-migration — and v1 rendering dies with JSON-RPC
 * in July 2026.
```

portal/common/lib/src/redirects.ts
```diff
@@ -33,19 +33,21 @@ export function redirectToAggregatorUrlResponse(blobId: string, aggregatorUrl: s
...
-    if (object.data && object.data.display) {
-        let display = object.data.display;
-        // Check if "walrus site address" is set in the display field.
-        if (display.data && display.data["walrus site address"]) {
-            return display.data["walrus site address"];
-        }
-    }
-    return null;
+    // Check if "walrus site address" is set in the rendered Display fields.
+    const address = object.display?.output?.["walrus site address"];
```
ioannischtz, line 49: Optional: a partially failed v2 render (`display.errors` non-null, key absent from `output`) is indistinguishable here from "not a redirector". A one-line `logger.warn` when `object.display?.errors` is non-null would separate render failures from plain sites.

(Parity note: JSON-RPC's `display.error` was equally ignored — no regression either way.)

portal/common/lib/src/url_fetcher.ts
ioannischtz, line 134: (reported by Claude, maybe relevant to SEW-1035 ?)

**`routingPromise` is fire-and-forget, and `getRoutesAndRedirects` can still *reject*** (RPC exhaustion; the malformed-DF `throw`). Every terminal fetch result — `Ok`, `BlobUnavailable`, `AggregatorFail`, `HashMismatch` — returns without awaiting it, so a rejection becomes an unhandled promise rejection: log noise on Bun / the service worker, fatal under Node semantics.

Pre-existing, but this PR built the miss channel, so flagging here. Cheapest fix is a `.catch` at creation that folds into the existing channel:

```ts
const routingPromise = this.wsRouter.getRoutesAndRedirects(resolvedObjectId).catch((e) => {
    const err = e instanceof Error ? e : new Error(String(e));
    return { routes: err, redirects: err };
});
```

…which is also the direction the `parseDynamicFieldValue` TODO already points.

portal/common/lib/tests/resource.test.ts
```diff
@@ -109,47 +86,13 @@ describe("fetchResource", () => {
...
-        };
+        const mockObject = mockSiteObject({
+            display: { output: { "walrus site address": "mockAddress" }, errors: null },
+        });
+
+        const mockResource = mockSiteObject({
+            display: { output: { "walrus site address": "mockAddress" }, errors: null },
+        });
 
         (checkRedirect as any).mockResolvedValueOnce(undefined);
```
ioannischtz: (From Claude)

**`checkRedirect` is sync, but this stub returns a Promise — which is truthy.** The redirect branch fires with a *Promise as the objectId*, so the recursion these assertions observe is an accident of truthiness, not the display-based redirect. Same pattern at L120 and L136.

The `display` fixtures on `mockObject`/`mockResource` are dead weight — the module mock means `display` is never read.

Suggest: `mockReturnValueOnce("0xredirectTarget")` / default `mockReturnValue(null)`, drop the display fixtures, and assert the recursion actually targets `"0xredirectTarget"` (e.g. via the second `multiGetObjects` call's args).

portal/common/lib/tests/resource.test.ts
```diff
@@ -176,23 +119,9 @@ describe("fetchResource", () => {
...
-                        version: "1.0",
-                    },
-                    digest: "",
-                    objectId: "",
-                    version: "1.0",
-                },
-            },
-            {},
+            mockSiteObject({ version: "1.0" }),
+            new Error("Object 0xdynamicFieldId not found"),
```
ioannischtz: (Claude)

**`"Object 0xdynamicFieldId not found"` does not match `isObjectNotFoundError`** — `y` isn't hex; the regex requires `0x[0-9a-f]+ not found`.

Harmless while resource.ts gates on bare `instanceof Error`, but routing.test.ts deliberately pins its mock misses with `expect(isObjectNotFoundError(mock)).toBe(true)` so mocks can't drift from the classifier. Same here: hex ids (`"Object 0xdf not found"`) plus that assertion. Also L141.

portal/common/lib/tests/grpc_get_objects_shape.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import { isObjectNotFoundError } from "@lib/rpc_selector";
+
+/**
+ * Drift guard for the gRPC `core.getObjects` shape, checked against mainnet — the
+ * canonical, stable network (testnet is periodically reset).
+ *
+ * `RPCSelector.multiGetObjects` depends on a specific contract from the SDK +
+ * fullnode, and silently relies on it: for a batch of object ids, `getObjects`
+ * returns a single `{ objects: [...] }` array, in request order, where
+ *   - a MISSING object is an `Error` ELEMENT (not thrown, not omitted), which our
```
ioannischtz: (CLaude)

**Stale: nothing collapses misses to `null`.** The `Error` elements pass through `multiGetObjects` unchanged — resource → 404, routing → `Routes | Error`; only `getNameRecord` resolves to `null`. Same phrase at L55–57 ("our Error→null mapping"). Worth fixing so the guard's failure message points at the real mapping.

FYI while here: upstream builds these miss elements as `new Error(message)` with **no `code`** and a `// TODO: improve error handling` (ts-sdks `grpc/core.ts`), and the message text comes from the fullnode, not the SDK. If upstream ever attaches a `code`, the `code`-shield in `isObjectNotFoundError` starts rejecting authoritative misses — and this guard goes red exactly then. Live-network pinning is the right call.

portal/common/lib/tests/grpc_get_objects_shape.test.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Requires MAINNET_RPC_URL (set via .env.test). We error rather than skip when
+ * it's absent, so this guard can't silently vanish.
+ */
+const mainnetRpc = process.env.MAINNET_RPC_URL;
+if (!mainnetRpc) {
+    throw new Error(
+        "MAINNET_RPC_URL must be set to run the gRPC drift guard (normally provided by .env.test)",
+    );
+}
+const MISSING_OBJECT_ID = "0x" + "0".repeat(63) + "f"; // valid-shaped id, never exists
```
ioannischtz: `0x…0f` sits in the reserved low-address space — `0xd` was just claimed by the Display registry, so "never exists" erodes with framework releases. A full-width improbable id is a safer forever-miss:

```suggestion
const MISSING_OBJECT_ID = "0x" + "deadbeef".repeat(8); // valid-shaped id, never exists
```

portal/common/.env.test
```diff
@@ -1,5 +1,7 @@
 # Format: URL|RETRIES|METRIC (lower metric number = higher priority)
 RPC_URL_LIST="https://fullnode.testnet.sui.io|2|100,https://sui-testnet.public.blastapi.io|1|200,https://testnet.suiet.app|1|300"
```
ioannischtz, line 2: Are the `blastapi.io` / `suiet.app` entries gRPC-capable? If they only ever spoke JSON-RPC, these two failover entries are dead config for the network tests — they'd surface only when `fullnode.testnet` is down, after burning their retries.

If there's no known gRPC failover for testnet, a single-entry list is more honest.

portal/worker/package.json
```diff
@@ -3,7 +3,7 @@
     "description": "The service-worker based implementation of the Walrus Sites portal.",
     "version": "2.10.0",
     "dependencies": {
-        "@mysten/sui": "^2.8.0",
+        "@mysten/sui": "^2.20.1",
```
ioannischtz, line 6: (Claude)

**Release-verification ask:** with this bump the service-worker portal issues gRPC-web requests *from the browser* — preflighted POSTs with a grpc-web content type.

Worth one manual in-browser run of the worker portal against a real fullnode before release. Fullnode CORS/preflight support is the failure mode nothing in CI exercises.

ioannischtz (commented):

One final coment from Claude regarding the deps that seems worth an extra look: `suins bump pulls axios 1.18.1 (CVE-positive). `

portal/common/lib/src/resource.ts
nikos-terzo: I do not know if we should actually keep this behavior. I just didn't want to do any behavior changes in the scope of this PR. Just added the note, when I was self reviewing the code. Maybe it makes sense to mark it as TODO: Decide on whether we want this.

portal/common/.env.test
nikos-terzo: I actually tried that, and they return unauthorized. This is the same, even before with json-rpc, so I didn't want to touch this line.
My assumption is that they use the sui repo behind the scenes for serving the FN, so I expect them to support grpc.

portal/common/lib/src/rpc_selector.ts
nikos-terzo: Agreed — but the subclass is pre-existing; I tried to keep this PR's scope as minimal as possible, so leaving the $extend migration for a follow-up.

portal/common/lib/src/rpc_selector.ts
nikos-terzo: `url`/`getURL()` are pre-existing too — kept them untouched to keep the PR's scope as minimal as possible.

portal/common/lib/src/rpc_selector.ts
nikos-terzo: Addressed at 25a06cf — added error_classifiers.test.ts with the suggested false-positive pins (code-shield, non-hex id, ^ anchor, non-Error inputs; kept the `i` flag and pinned it as intended).

portal/common/lib/src/rpc_selector.ts
nikos-terzo: Addressed at 1b97f22 — `code === "INVALID_ARGUMENT"` now logs as error and returns `stop`, so a malformed request aborts on the first attempt instead of sweeping the failover list (unit test: 1 call instead of 6). Other codes still retry/fail over since they can differ per node.

ioannischtz (approved):
pre approving, ty @nikos-terzo ✅

portal/common/lib/src/rpc_selector.ts
nikos-terzo: Follow-up: chasing this class of error also caught a pre-existing bug — SuinsClient throws `Invalid SuiNS name` pre-network for a malformed (user-typed) subdomain, which swept the whole failover list and surfaced as the 503 "failed to contact the full node" page (and inflated the FN-failure metric, likely feeding SEW-979's false alerts). Fixed at 552b5ed: validate once up front and resolve to null (site not found). UX follow-up (dedicated "check for typos" message) tracked in SEW-1036.

portal/common/lib/src/resource.ts
nikos-terzo: Addressed at e798110 — info log on a site-object miss, plus a warn when the site object is missing but the resource still resolves (i.e. we're serving a deleted site's orphaned resource — whether we should keep doing that is now tracked in SEW-1037, with a test pinning current behavior until decided).

portal/common/lib/src/redirects.ts
nikos-terzo: Agreed, but keeping it out of this PR's scope — tracked in SEW-1038. Going with `info` rather than warn: the Display template belongs to a third party (the redirecting object's package author), so it's their bug to fix — we just want visibility in the logs.

portal/common/lib/src/redirects.ts
nikos-terzo: One correction to the exposure analysis: the redirect Display isn't (only) `Display<Site>` — `checkRedirect` runs on whatever object the subdomain resolved to, which in the NFT-redirect pattern is a third-party object (e.g. a Flatlander), whose Display belongs to that package's author. So verifying mainnet `Display<Site>` renders doesn't bound the risk: any third-party author who created their Display via the v1 API after the snapshot and hasn't migrated will silently lose their redirect over gRPC. Still the same conclusion (v1 dies with JSON-RPC in July 2026 anyway), but the affected Displays are ones we don't control.

portal/common/lib/src/url_fetcher.ts
nikos-terzo: Confirmed — this is exactly SEW-1035, which I filed from the e2e parity testing: reproduced identically on main (JSON-RPC v2.10.0) and on this branch, so it's pre-existing, not a regression of the migration. Keeping the fix out of this PR's scope; the `.catch`-at-creation approach suggested here matches the ticket's expected fix.

portal/common/lib/tests/resource.test.ts
nikos-terzo: Fixed at 39c8cc0 — and the enabler for this bug was the `(checkRedirect as any)` cast: with a typed `vi.mocked(checkRedirect)` stub, calling `mockResolvedValueOnce` on a sync function returning `string | null` doesn't even compile. Switched all stubs in the file to `vi.mocked`, redirect case now returns `"0xredirectTarget"`, dead display fixtures dropped, and the test asserts the recursion targets the redirect id via the second `multiGetObjects` call.

portal/common/lib/tests/resource.test.ts
nikos-terzo: Fixed at 0e66681 — mock misses now go through a `mockObjectMiss()` helper that uses hex ids and asserts `isObjectNotFoundError(miss)` inline, same pattern as routing.test.ts, so they can't drift from the classifier again.

portal/common/lib/tests/grpc_get_objects_shape.test.ts
nikos-terzo: Right, stale — fixed at ceada44: the doc and failure messages now point at the real mappings (resource → 404, routing → `Routes | Error`, only getNameRecord → null). Noted on the upstream FYI — the live-network pinning stays, so if ts-sdks ever attaches a `code` to miss elements this guard goes red exactly then.

portal/common/lib/tests/grpc_get_objects_shape.test.ts
nikos-terzo: Good point — applied at 678cd32 (`"0x" + "deadbeef".repeat(8)`), verified against mainnet.

portal/worker/package.json
nikos-terzo: Done — tested manually in-browser against mainnet (worker built with mainnet config, served locally, with the diff from SEW-1039 applied). Result: browser gRPC-web works — preflighted POSTs to the fullnode are accepted (`access-control-allow-origin: *`, `allow-headers: *`) and docs.sui resolves and renders. The worker did NOT boot out of the box, but for pre-existing reasons unrelated to this bump: the instrumentation module (bare `process` + Prometheus exporter at SW load) and the global `Buffer` in quilt.ts — both documented with the fix diff in SEW-1039, kept out of this PR's scope.

nikos-terzo:
On the axios 1.18.1 note: axios itself is the latest release — the CVE-positive pieces are its transitive deps. `follow-redirects` is already resolved (lockfile has the patched 1.16.0). `form-data` 4.0.5 is in-range for GHSA-hmw2-7cc7-3qxx but not exploitable from the portal: axios is only used by suins's Pyth price-feed module (name-purchase flow), which the portal never calls, and the CVE needs attacker-controlled multipart field names — the portal sends no multipart anywhere. We'll pick up the fixed resolution the moment a newer suins SDK lands. Also filed SEW-1040 to run `bun audit` in CI so these surface on PRs instead of in review.

PR #725, opened 2026-07-09
fix(site-builder): migrate off JSON-RPC-only client methods and bump walrus deps to testnet-v1.51.2

shuowang12:
## Context

Sui testnet fullnodes no longer serve JSON-RPC. Walrus testnet-v1.51.2 fixes client startup against gRPC-only endpoints and makes its remaining JSON-RPC-only wrapper methods crate-private (they can't work against testnet at runtime regardless). That visibility change is why the automated bump to v1.51.2 failed on Jul 9 with two E0624 errors: site.rs calls two of those wrappers. This PR replaces the automated bump.

That visibility change is what broke the automated dependency bump here ("Bump dependency versions" run on Jul 9, two E0624 errors): site-builder calls two of those wrappers in `site.rs`. This PR supersedes the failed automated bump and unblocks the site-builder release, which is the last leg before the walrus scheduled-compatibility canary can go green again.

## Changes (two commits, lockstep)

**1. Migrate the two site.rs call sites to gRPC-backed APIs**

 - Resource dynamic fields: multi_get_object_with_options_batched + manual deserialization → get_sui_objects::<ResourceDynamicField>(). Drop-in: it batch-fetches and BCS-decodes, chunking internally at 100 objects per request.
  - Routes/redirects: same DF-id derivation as before, but existence is checked against the site's already-fetched dynamic-field listing (no extra RPC), then each existing field is fetched with get_sui_object. This is the part worth reviewer attention: it preserves missing-field ⇒ None without mapping fetch errors to None — relevant because SiteData feeds the update diff, and a transient error misread as "no routes" could delete a site's routes on update.

**2. `chore(site-builder): bump walrus deps to testnet-v1.51.2`**

- Bumps all walrus pins from `testnet-v1.51.1` to `testnet-v1.51.2` plus the lockfile. The Sui tag stays at `testnet-v1.74.0`, which already matches walrus v1.51.2.

The two changes only work together: the migration alone still fails E2E (v1.51.1 crates cannot construct a Sui client against testnet), and the bump alone does not compile (the E0624s). Hence one PR, two commits for review clarity.

## Validation

- Compiles against both testnet-v1.51.1 and testnet-v1.51.2 pins.
- walrus v1.51.2's client path is already running in production on the testnet fleet.

## After merge
Cut a site-builder release so the published site-builder-testnet-latest binaries work against testnet again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1562,8 +1562,8 @@ dependencies = [
 
 [[package]]
 name = "checkpoint-downloader"
-version = "1.51.1"
-source = "git+https://github.com/MystenLabs/walrus?tag=testnet-v1.51.1#01ada87c9f40e8bde9ad81861a8fbe6b20878361"
+version = "1.51.2"
+source = "git+https://github.com/MystenLabs/walrus?tag=testnet-v1.51.2#7741b19a32941912f0a46ac34bf6b05ede86f3c3"
 dependencies = [
  "anyhow",
  "async-channel",
@@ -1579,7 +1579,7 @@ dependencies = [
  "tokio",
  "tokio-util 0.7.18 (registry+https://github.com/rust-lang/crates.io-index)",
  "tracing",
- "typed-store 1.51.1",
+ "typed-store 1.51.2",
  "walrus-sui",
  "walrus-utils",
 ]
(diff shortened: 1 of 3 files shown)


shuowang12:
## Why "E2E: Upgrade Test" is red and cannot go green on this PR

The check requires a working main, and main only starts working when this merges.  This makes the check structurally unable to pass for the PR that fixes main.

Requesting admin merge.

ebmifa (approved):

